PR #7, opened 2017-03-13
Use lowercase log-level

arledesma:
Vault is case sensitive in regards to log-level.  Passing 'INFO' is not the same as 'info'.

* Only append -log-level if variable is defined
* Default to lowercase info
* lowercase any vault_log_level passed into a template
* Allowed log-levels are not documented in official documentation, only in code.
  https://github.com/hashicorp/vault/blob/b1ed578f3da3263ca1973d16dcb33490125486b8/command/server.go#L1003-L1005


arledesma:
Further code documentation for `-log-level`

Defaulted to `info` if not specified
https://github.com/hashicorp/vault/blob/b1ed578f3da3263ca1973d16dcb33490125486b8/command/server.go#L73

Switch with error for unknown log-level
https://github.com/hashicorp/vault/blob/b1ed578f3da3263ca1973d16dcb33490125486b8/command/server.go#L87-L103

PR #8, opened 2017-03-14
Adds support for TLS

arledesma:
Should be backwards compatible.  

You may want to verify the changes locally.


brianshumate:
Thanks!

PR #12, opened 2017-03-23
TLS: Quote mode in with_items

arledesma:
The mode is being interpreted as decimal rather than octal when expanded.
Quoting ensures that the mode is no longer cast.


brianshumate:
Aha, so err on the side of *quote everything*  😰

PR #16, opened 2017-04-27
[fix] Add TLS directory if TLS is enabled

cwill747:
When TLS is enabled, the directory for the TLS cert is not created by default. This adds a task to tls.yml to create the directory that the TLS certs are written to.


brianshumate:
Looks good, thanks!

PR #20, opened 2017-07-27
Fix Missing Defaults for TLS

marc-sensenich:
Given that `vault_tls_min_version` and `vault_tls_prefer_server_cipher_suites` will evaluate to false, make sure the defaults are populated


brianshumate:
Thanks!

PR #21, opened 2017-08-08
Add missing redirect_addr in HA consul config

groggemans:
Redirect address is required for HA setup with consul.


brianshumate:
Thanks!

PR #23, opened 2017-08-29
add missing restart case in the debian init script

Gerrrr:
Hello!

Thanks for this nice role!
I tried to use it with Ubuntu 14.04 and found a missing `restart)` case in the Debian init script:

Error message:

```
{"changed": false, "failed": true, "msg": "/etc/init.d/vault: 90: /etc/init.d/vault: Syntax error: word unexpected (expecting \")\")\n"}
```

After adding the restart case the role succeeds.


groggemans:
Thank you!

PR #26, opened 2017-09-01
Remove vault status check from the debian script

Gerrrr:
There are 2 problems with the `vault status` check in the debian init script right now:

* If Vault is being installed for the first time, it is sealed and `vault status` will return non-zero exit which in turn will make the init script return non-zero exit code
* With configured TLS calling `vault status` without specifying Vault's public address or domain name used in the certificate will end up in the certificate check error

Also, this check does not exist in the daemon scripts for the other platforms, so removing it unifies the behavior.

---
Sorry for multiple PRs to the same file changing the behavior, I hope it does not annoy you too much :)

### DIFF
--- a/templates/vault_debian.init.j2
+++ b/templates/vault_debian.init.j2
@@ -39,17 +39,12 @@ do_start() {
         $DAEMON_ARGS \
         || return 2
 
+    RETVAL=0
     for i in `seq 1 30`; do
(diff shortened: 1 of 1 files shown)


groggemans:
@Gerrrr  NP! All PR's are welcome :+1: 

I agree that removing the check altogether might be better.

PR #28, opened 2017-09-04
Fix default paths for tls cert and key files in README

Gerrrr:
Defaults:

* [vault_tls_cert_file_dest](https://github.com/brianshumate/ansible-vault/blob/master/defaults/main.yml#L66)
* [vault_tls_key_file_dest](https://github.com/brianshumate/ansible-vault/blob/master/defaults/main.yml#L70)


groggemans:
Thank you!

PR #35, opened 2018-03-20
Fix is Vault api reachable task

rarguelloF:
Fixed the task: "Vault API reachable?" for the following cases:

- `vault_address ` is set to `0.0.0.0` -> in this case, the request is made to `127.0.0.1`
- There was changes in the listen address in the configuration, in that case it would be necessary to restart vault first.
- Added a timeout of 1 minute.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -113,8 +113,12 @@
     state: started
     enabled: yes
 
+- name: Restart Vault if needed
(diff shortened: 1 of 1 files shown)


brianshumate:
Cool, thanks; I'll have a look!

brianshumate:
👍 Thanks!

PR #44, opened 2018-05-09
Add support for HTTP 429 when vault_cluster_disable is false. Fixes #39

### LINKED ISSUE #39: return code 429 is also valid in HA mode
In HA mode, only one vault node will become active node and therefore return status code 200 when queried for /sys/health. All other nodes will become standby and return 429 (unsealed and standby). Still, if return code 429 is returned, it means that node is unsealed and part of the "cluster". 

Would it make sense that task called "Vault API reachable? If not please unseal it" in tasks/main.yml accepts status_code 429 besides 200?

Rerefence: 
https://www.vaultproject.io/api/system/health.html
https://www.vaultproject.io/docs/concepts/ha.html

eripa:


brianshumate:
Thanks!

PR #46, opened 2018-05-09
Add support for disabling certificate validation during API reachability check. Fixes #45

### LINKED ISSUE #45: Add option to skip certificate validation in API reachability check
Currently in the task **Vault API reachable? If not please unseal it** there's no support for skipping certificate validation. This is problematic since the `vault_address` defaults to an IP address. It can also be useful if you rely on self-signed certs. 

I can submit a PR for this.

eripa:


brianshumate:
Thank you!

PR #47, opened 2018-05-22
allow for premium users to install with minimal changes

drewmullen:
adds +prem to vault_version var if vault_enterprise_premium=True

also, add handler to restart vault servers if listener config changes


drewmullen:
found a copy pasta error. closing, will reopen

PR #48, opened 2018-05-22
add +prem to vault_version if new var, vault_enterprise_premium, is true

drewmullen:
add +prem to vault_version if new var, vault_enterprise_premium, is true

also, fire restart handler when listener config changes


brianshumate:
Thanks!

PR #50, opened 2018-05-31
improve TLS features and add vault enterprise download option

drewmullen:
Added CA file location to TLS listener in vault config

Added ability to enable TLS gossip between vault and consul

Cleaned up spacing on config file entries

Added vault enterprise download when the zips are not found locally. This will allow internally hosting rather than placing the file on the ansible local host

Cleaned up some TLS var names for cleaner cert deployment. some file deployments were named as part of the path, this is confusing and easier to maintain when separated out.


drewmullen:
gotta include a download for the checksum, had the file in my local files/ dir

PR #51, opened 2018-06-01
improvements to TLS, encrypt gossip, enable hsm binary installs

drewmullen:
Added CA file location to TLS listener in vault config

Added ability to enable TLS gossip between vault and consul

Cleaned up spacing on config file entries

Added vault enterprise download when the zip (and checksum) is not found locally. This will allow internally hosting rather than placing the file on the ansible local host

Enable installation of hsm binaries using bool vault_enterprise_premium_hsm var

Cleaned up some TLS var names for cleaner cert deployment. some file deployments were named as part of the path, this is confusing and easier to maintain when separated out.


defaults/main.yml
```diff
@@ -51,7 +51,7 @@ vault_ui: "{{ lookup('env', 'VAULT_UI') | default(false, true) }}"
 vault_port: 8200
 vault_node_name: "{{ inventory_hostname_short }}"
 vault_main_config: "{{ vault_config_path }}/vault_main.hcl"
-vault_backend: vault_backend_consul.j2
+vault_backend_consul: vault_backend_consul.j2
```
brianshumate, line 54: This role currently supports both the Consul and File storage backends (and potentially others in the future); not sure this change makes sense.

templates/vault_listener.hcl.j2
```diff
@@ -17,8 +17,13 @@ listener "tcp" {
   tls_disable = "{{ vault_tls_disable | bool | lower }}"
 }
 
-{% include vault_backend with context %}
+{% include vault_backend_consul with context %}
```
brianshumate, line 20: Same as above. Please leave this as-is.

brianshumate (changes requested):
Thanks for the HSM bits here! I don't think we want to rename the `vault_backend` option though.

brianshumate:
Thank you!

drewmullen:
you decide its okay to keep those changes to the vault_backend name?

PR #57, opened 2018-09-06
Avoid unneeded installation tasks run

jpiron:
Don't run installation tasks when Vault is already installed with the
required version.
This is to prevent playbook runs to report unwanted changes.


brianshumate:
Cool, thanks!

PR #58, opened 2018-09-06
Add a condition on mlock capability addition

jpiron:
To prevent command from running and reporting changes when unneeded.


brianshumate:
Thank you!

PR #59, opened 2018-09-10
Fix consul backend template

jpiron:
vault_tls_gossip is evaluated as a string and is considered true even if
it holds 0.


brianshumate:
Thanks!

PR #64, opened 2018-10-16
Allow overriding of templates in tasks/main.yml

nathkn:
This commit sets all the templates referenced in the tasks/main.yml as
variables so that they can be overriden by the user, without needing to
update the roles directory.

Re-solves #2


brianshumate:
Nice, thank you!

PR #66, opened 2018-11-15
Enhanced role for file based storage backend and testing.

snoby:
Notes on changes:

* Removed unused vault_group_name variable
* Added an option for the playbook to create the os group if it doesn't exist
* Added an option for the playbook to initialize a vault instance, this is mainly for testing.
* Added storage backend for file, mainly used for testing without a consul cluster.
* Updated the vault_listener.hcl.j2 template fale to handle different storage backends, such as consul, file, and mysql
* Changed the end of the playbook check for if the API is reachable.  When using a brand new cluster that has never been initialized the check to see if it's initialized will always work if the API is reachable.  If not initialized the heath endpoint will not respond.
* Added a check to see if the vault has been initialized
* Added ability to initialize the vault and print out the initialization keys to stdout, this is mainly used for testing.
* Changed the API check to check the health endpoint only if the vault has been initialized.
* Changed the url used to reach the vault instance by surrounding the colons in {{':'}} because otherwise the jinja template engine in version 2.7 of ansible barfs on the colon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,10 +218,20 @@ The role defines variables in `defaults/main.yml`:
 - Vault listener configuration template file
 - Default value: *vault_listener.hcl.j2*
 
+### `vault_backend`
+- Which storage backend should be selected, choices are: consul, file and mysql
+- Default value: *consul*
+
 ### `vault_backend_consul`
 
 - Backend consul template filename
-- Default value: `backend_consul.j2`
+- Default value: *backend_consul.j2*
+
+### `vault_backend_file`
+
+- Backend file template filename
+- Default value: *backend_file.j2*
+
 
 ### `vault_cluster_address`
 
@@ -491,6 +501,22 @@ role directory for this to work.
 
 See `examples/README_VAGRANT.md` for details on quick Vagrant deployments
 under VirtualBox for testing, etc.
+
+## example virtualBox playbook
+example playbook for a file based  vault instance.
+
+```
+- hosts: all
+  gather_facts: True
+  become: true
+  vars:
+    vault_backend: file
+    vault_cluster_disable: True
+    vault_log_level: debug
+  roles:
+    - vault
+
+```
 
 ## Vault Enterprise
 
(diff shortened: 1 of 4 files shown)


brianshumate:
Hey there,

Thanks for the PR. I think there are probably too many unrelated changes in this single PR and would prefer to see it as a series of PRs with limited specific changes if you don't mind.

For example:

> Removed unused vault_group_name variable

This is not related to the other changes and probably should be a separate PR.

> Added storage backend for file, mainly used for testing without a consul cluster.
> Updated the vault_listener.hcl.j2 template fale to handle different storage backends, such as consul, file, and mysql

Using the filesystem storage is already possible, but not yet documented.

There's a [vault_backend_file.j2](https://github.com/brianshumate/ansible-vault/blob/master/templates/vault_backend_file.j2) template for doing so that can be used by:

1. Defining the preferred filesystem path (defaults to `/var/vault`) in the `vault_data_path` variable
2. Specifying `vault_backend_file.j2` as the value to the `vault_backend_consul` variable 

I can add some notes to the README about using filesystem based storage.

> Added a check to see if the vault has been initialized
> Added ability to initialize the vault and print out the initialization keys to stdout, this is mainly used for testing.
> Changed the API check to check the health endpoint only if the vault has been initialized.

The role is focused on initial deployment of Vault, and this deliberately does not include initialization, unsealing, or post-unseal configuration. Any work towards automating those aspects of Vault configuration need to occur outside of this role.

> If not initialized the heath endpoint will not respond.

This is incorrect; the [/sys/health](https://www.vaultproject.io/api/system/health.html) endpoint is unauthenticated, and always available when Vault is operational regardless of initialization state.

For example, this response is from a running but uninitialized Vault server `/sys/health` API:

```
curl -s https://localhost:8200/v1/sys/health | jq .
{
  "initialized": false,
  "sealed": true,
  "standby": true,
  "performance_standby": false,
  "replication_performance_mode": "unknown",
  "replication_dr_mode": "unknown",
  "server_time_utc": 1542302674,
  "version": "0.11.4"
}
```

Thanks again, but we'll need to reconsider what items from your work we can accept as PRs and also try reduce their scope to smaller/more specific ones going forward.

brianshumate:
Going to close this since we merged the PR. Thanks again!

PR #68, opened 2018-11-15
fix: use AmbientCapabilities only when systemd version is >= 230

### LINKED ISSUE #67: Vault wont start due to AmbientCapabilities in Systemd unit file
In version v2.0.7 of this role, the systemd unit file includes the following line:

AmbientCapabilities=CAP_SYSLOG CAP_IPC_LOCK

However, on a Centos7 host running Linux 3.10.0-693.11.1.el7.x86_64 with systemd 219, this causes vault to fail to start with the following log message:

    systemd[20617]: Failed at step CAPABILITIES spawning /usr/local/bin/vault: Operation not permitted

Removing this line allows systemd to start.

It sounds to me like the playbook should only attempt to include that line if the software version supports it, perhaps this should be wrapped in a conditional check instead of being included on every system. 

Note: Yum shows no pending upgrades available for this system, so I don't think this can be solved with a yum upgrade.

bdossantos:
Hi,

The `AmbientCapabilities` directive I've introduced to the systemd config in brianshumate/ansible-vault#65 cause a breakage on "old" systemd versions. 
This patch extract version and include `AmbientCapabilities` directive only if version is `>=` to `230`

Sorry for the breakage :-/

close: brianshumate/ansible-vault#67

### DIFF
--- a/templates/vault_systemd.service.j2
+++ b/templates/vault_systemd.service.j2
@@ -25,7 +25,9 @@ PrivateTmp=yes
 PrivateDevices=yes
 SecureBits=keep-caps
 Capabilities=CAP_IPC_LOCK+ep
+{% if systemd_version | version_compare('230', '>=') %}
 AmbientCapabilities=CAP_SYSLOG CAP_IPC_LOCK
+{% endif %}
 CapabilityBoundingSet=CAP_SYSLOG CAP_IPC_LOCK
 NoNewPrivileges=yes
 ExecStart={{ vault_bin_path }}/vault server -config={{ vault_main_config }} {% if vault_log_level is defined %}-log-level={{ vault_log_level | lower }}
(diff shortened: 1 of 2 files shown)


templates/vault_systemd.service.j2
```diff
@@ -25,7 +25,9 @@ PrivateTmp=yes
 PrivateDevices=yes
 SecureBits=keep-caps
 Capabilities=CAP_IPC_LOCK+ep
+{% if systemd_version | version_compare('230', '>=') %}
```
bdossantos, line 28: It may be safer to cast this variable to int, but I don't know how to do it when registering it https://github.com/brianshumate/ansible-vault/pull/68/files#diff-2444ad0870f91f17ca6c2a5e96b26823R139

brianshumate:
👋 yes I've seen some differing issues around the capabilities based on systemd/distro version and it seems we do need a conditional like this so thanks much for bringing it up!

I'm not finding a lot of examples that cast the version value to an int in similar cases, so I think the `version_compare` alone should suffice for now. Thanks for doing this change!

PR #80, opened 2019-02-01
use a variable to skip the copy of certificate

ghost:
Like in the consul role, let the certificates to be copied in a different way


brianshumate:
Thanks!

PR #81, opened 2019-02-04
variable to skip the healthcheck, useful when vault is not initialized

ghost:
Use a variable to skip the healthcheck, useful for the first run when vault is not yet initialized


brianshumate:
Thank you!

PR #84, opened 2019-02-07
Added additional status codes to Vault API check.

### LINKED ISSUE #83: Vault API reachable? - fails
Hello, I'm sad to say but I can't get this role to successfully complete the `Vault API reachable?` task when running the example locally in Vagrant: https://github.com/brianshumate/ansible-vault/blob/master/tasks/main.yml#L179.

All the previous tasks seem to complete just fine when running `vagrant up` from the examples directory, but the `uri` module is cycling for 30 iterations without a response at http://10.1.42.240:8200/v1/sys/health and then exiting with an error on that task.

I have setup `/etc/hosts` and the vagrant-hosts plugin, and  I even tried spinning up the consul role from https://github.com/brianshumate/ansible-consul to see if it was an expected dependency but no luck. I noticed a few minor typos in the readme: https://github.com/brianshumate/ansible-vault/blob/master/examples/README_VAGRANT.md. Is there a step I may be missing?

Thanks!

enqack:
Fixes #83


brianshumate:
Thanks

PR #86, opened 2019-03-20
Fix install failure when .bashrc does not exists

gfeun:
All users do not have a .bashrc. 
[PR76](https://github.com/brianshumate/ansible-vault/pull/76) adds lines to bashrc but causes install failure if the file does not exist.

Solution: allow bashrc file creation if it does not exist.


brianshumate:
Thanks!

PR #90, opened 2019-04-10
Add vault_bin_path to the PATH

kwevers:
The vault_bin_path might not be included in the search path
for the user executing Ansible.
This will cause the install to be done over and over again.

At least on systems running RHEL the, default, /usr/local/bin
vault_bin_path is not included in the PATH of root


brianshumate:
Thanks!

PR #91, opened 2019-04-10
Ensure systemd is reloaded on unit changes

kwevers:
If systemd doesn't get reloaded it will issue a warning and use the
old service definition


brianshumate:
Nice, thank you!

PR #92, opened 2019-04-10
Revert long line syntax change

kwevers:
Using the folded block scalar causes a space to be added between the scheme and the colon

This results in a line looking like this:
```
"export VAULT_ADDR='http" "://localhost.local:8200'"
```
Causing the below error when sourcing .bashrc
```
-bash: export VAULT_ADDR='http: command not found
```


brianshumate:
😓  Thanks for catching this!

PR #93, opened 2019-04-10
Vault health check

kwevers:
- Ensure the health check also succeeds when the vault is sealed for a standalone node
- Remove obsolete `vault_health_check` variable


brianshumate:
Thank you!

PR #94, opened 2019-04-11
Remove if construct for vault_log_level

kwevers:
Using the default filter in defaults/main.yml will cause this variable to always have a value.

The bool filter would always return false unless somebody sets the
vault_log_level to true/yes/.. removing the `-log-level` statement from the result altogether.

Example:
```
ExecStart=/usr/local/bin/vault server -config=/etc/vault/vault_main.hcl ExecReload=/bin/kill --signal HUP $MAINPID
```


brianshumate:
Thanks!

PR #95, opened 2019-04-18
add etcd storage for vault

cordula-grau:
* add etcd storage for vault
* resolve deprecation warnings regarding version_compare
* add become flags on task level to avoid become flag on role inclusion


brianshumate:
Thank you!

PR #97, opened 2019-05-10
Add plugin path configuration

vmwiz:
I need the ability to load custom plugins in my vault instance, added a new variable `vault_plugin_path` and proper vault configuration parameter to enable plugins usage.


brianshumate:
Awesome, thanks!

PR #102, opened 2019-06-13
Add S3 backend

dvmonroe:
Variables and defaults are based on [documentation](https://www.vaultproject.io/docs/configuration/storage/s3.html)

Also removes mysql from README since it's not yet supported by this role


brianshumate:
Thanks!

PR #103, opened 2019-06-13
[Documentation] Fix typo and update README

dvmonroe:
I could be wrong about the intention on the README...


brianshumate:
Good catch and thank you!

PR #104, opened 2019-06-13
Fix template line break

fhemberger:
Newline was missing before "ui":

```
storage "file" {
  path = "/var/vault"
}ui = true
```


fhemberger:
Hi, would be cool if you find the time to merge this. Thanks!

brianshumate:
Thanks!

PR #105, opened 2019-06-13
Use 'ansible_default_ipv4' interface as default

fhemberger:
The network interface 'eth1' does not always exist, so instead fall back to
`hostvars[inventory_hostname]['ansible_default_ipv4']['interface']`.


fhemberger:
Hi, would be cool if you find the time to merge this. Thanks!

brianshumate:
Thanks!

PR #109, opened 2019-07-08
Add support for backend tls configuration

rhenwood3995:
This pull request adds variables for setting backend tls options. Previously, they were hardcoded to the listener tls options. The new backend options default to the listener options.


brianshumate:
Thanks!

PR #111, opened 2019-07-18
Fix multiple issues

zeridon:
* vault_home is wrong -> move to vault_user. The need for this is only to add some vars (which happening on the remoe machines is kind of not needed)
 * change the logic for vault_iface to get the default ansible interface and not statically eth1


brianshumate:
Thanks!

PR #112, opened 2019-07-29
Add support for AWS DynamoDB storage backend

chris-dudley:
Adds support for configuring Vault to use the DynamoDB storage engine.

Documentation for all of the backend-specific variables and their defaults have been added to the README. The defaults are based on the AWS KMS unseal backend defaults, allowing for
overriding of most of the defaults via environment variables.

Note: This PR also includes some changes to the AWS KMS template as well, as I ran into issues
with that template while testing a full deployment with my changes.
I was having issues with values not appearing in the generated configuration file due to the existing `if var | bool` checks.
If you would prefer to deal with those changes separately, I can back them out of this PR and submit them in a separate PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ The role defines variables in `defaults/main.yml`:
 ## Storage Backend Variables
 
 ### `vault_backend`
-- Which storage backend should be selected, choices are: consul, etcd, file, and s3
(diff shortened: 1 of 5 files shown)


brianshumate:
Thanks!

PR #113, opened 2019-08-01
Fix Vault installation check

jpiron:
Command require -v option to output the binary path.
Also, command seems to require a shell to work.


brianshumate:
Thanks!

PR #115, opened 2019-09-11
fixes incorrect template syntax and adds the template for vault_backend_s3

ebostijancic:
Using s3 as vault backend was failing, this fixes it.


brianshumate:
Thanks!

PR #116, opened 2019-09-17
🐛 notify: Restart vault, fix for #108

bbaassssiiee:
Fix for #108
After installation of the new vault binary, restart.


brianshumate:
Thanks!

PR #126, opened 2019-11-15
added mysql storage option that was missing

ericsysmin:
As noted there was a missing MySQL storage file. This one has been tested and supports all options.


brianshumate:
Thanks

Issue #124, opened 2019-11-15
"msg": "hostname '10.138.0.95' doesn't match 'vault.ericsysmin.local'",


ericsysmin:
https://github.com/brianshumate/ansible-vault/blob/d3f4353c5ed9a1da120cc484b8c3ad106de9099f/tasks/main.yml#L238

This line here doesn't allow you to use a valid domain name so it will fail due to validate certs on https.

bbaassssiiee:
https://github.com/ansible-community/ansible-vault/pull/128 was merged

Issue #122, opened 2019-10-15
api_addr should match TLS certificate


bbaassssiiee:
` /etc/vault.d/vault_main.hcl` has an IP address for api_addr, this is not included in my certificate.
This causes a TLS problem when I try using [Seth Vargo's Plugin](https://github.com/sethvargo/vault-secrets-gen) 

```
api_addr = "https://10.1.1.64:8200"
```

I would rather use "{{ ansible_fqdn }}" or "{{ inventory_hostname }}". Why not?

ericsysmin:
I am working on a 2nd pr that should fix this as it causes issues for me as well

ericsysmin:
@bbaassssiiee does that PR help?

bbaassssiiee:
Yes it does, thanks!

PR #128, opened 2019-11-15
added hostname support

ericsysmin:
This change was made to address #122 and #124

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -53,6 +53,7 @@ vault_datacenter: dc1
 vault_log_level: "{{ lookup('env','VAULT_LOG_LEVEL') | default('info', true) }}"
 vault_iface: "{{ lookup('env','VAULT_IFACE') | default(ansible_default_ipv4.interface, true) }}"
 vault_address: "{{ hostvars[inventory_hostname]['ansible_'+vault_iface]['ipv4']['address'] }}"
+vault_hostname: "{{ vault_address }}"
 vault_ui: "{{ lookup('env', 'VAULT_UI') | default(true, true) }}"
 vault_port: 8200
 vault_main_config: "{{ vault_config_path }}/vault_main.hcl"
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -53,7 +53,6 @@ vault_datacenter: dc1
 vault_log_level: "{{ lookup('env','VAULT_LOG_LEVEL') | default('info', true) }}"
 vault_iface: "{{ lookup('env','VAULT_IFACE') | default(ansible_default_ipv4.interface, true) }}"
 vault_address: "{{ hostvars[inventory_hostname]['ansible_'+vault_iface]['ipv4']['address'] }}"
-vault_hostname: "{{ vault_address }}"
 vault_ui: "{{ lookup('env', 'VAULT_UI') | default(true, true) }}"
 vault_port: 8200
 vault_main_config: "{{ vault_config_path }}/vault_main.hcl"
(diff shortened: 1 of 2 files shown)


defaults/main.yml
```diff
@@ -53,6 +53,7 @@ vault_datacenter: dc1
 vault_log_level: "{{ lookup('env','VAULT_LOG_LEVEL') | default('info', true) }}"
 vault_iface: "{{ lookup('env','VAULT_IFACE') | default(ansible_default_ipv4.interface, true) }}"
 vault_address: "{{ hostvars[inventory_hostname]['ansible_'+vault_iface]['ipv4']['address'] }}"
+vault_hostname: "{{ vault_address }}"
```
bbaassssiiee: Why not like this?
```
vault_hostname: "{{ inventory_hostname }}"
```

ericsysmin: because this still has to work even if you aren't using a hostname, and for backward compatibility. What if inventory_hostname is not a valid fqdn? If your inventory_hostname is a valid fqdn then you could provide that as a var to the role and it would work as desired.
@bbaassssiiee did that explanation make sense?

bbaassssiiee: Would this work?

`vault_hostname: "{{ ansible_fqdn|vault_address }}"`

ericsysmin: So, ansible_fqdn is always assigned via Ansible regardless of being resolvable. So it would be meaningless to have vault_address there. It also would break existing deployments using this role as it would change where the endpoint is expected. If a host has multiple IP addresses and you're hosting this one one of the IP's assuming ansible_fqdn can cause the wrong IP to be returned, ending up in a failed check, even if the service is running.

lets say my ansible_fqdn is ericsysmin, which resolves to 127.0.0.1 but my Vault service is running on 10.0.0.120, and it uses a different domain. How would you handle this?

Prior to this PR we were already using the IP address, so we need to ensure that existing deployments don't get broken because of DNS resolving to different addresses other than the vault_address which could have been set or assigned to the vault server.

ericsysmin: However, you did just help me realize a problem on checking when using 0.0.0.0 which uses all IP's on the host

bbaassssiiee (approved):

defaults/main.yml
ericsysmin: Removed this from default/main.yml

ericsysmin:
@bbaassssiiee here is an example of how I was using it on my lab
```yaml
- include_role:
    name: ericsysmin.vault
  vars:
    vault_version: 1.2.3
    vault_user: vault
    vault_group: vault
    vault_backend: mysql
    vault_address: 0.0.0.0
    vault_hostname: vault.ericsysmin.com
    vault_mysql: 172.27.1.2
    vault_mysql_username: vault
    vault_mysql_password: "{{ vault_db_password }}"
    vault_mysql_database: vault
    vault_max_lease_ttl: 10h
    vault_default_lease_ttl: 10h
    vault_tls_copy_keys: false
    vault_tls_disable: false
    vault_tls_config_path: /etc/letsencrypt/live/vault.ericsysmin.com
    vault_tls_cert_file: fullchain.pem
    vault_tls_key_file: privkey.pem
```

Originally I didn't have `vault_address` set which caused it to use my local ip from Ansible, which wouldn't catch the entire issue that if someone did set it to 0.0.0.0 it would have caused it not to work if vault_hostname wasn't set..

bbaassssiiee:
I'm working on this https://github.com/dockpack/vault_dojo

ericsysmin:
> I'm working on this https://github.com/dockpack/vault_dojo

Got it, nice example for deployment with Consul and Vault

ericsysmin:
@brianshumate can you merge this PR?

bbaassssiiee:
Thanks!

PR #129, opened 2019-11-15
simplified the message to not require loop

ericsysmin:
This should fix ansible-lint issue on jinja in when statement and speed up deployment.


brianshumate:
Thanks

Issue #89, opened 2019-04-09
Clarification required about vault_checksum_file_url variable content for Enterprise Install


seb-braun:
While deploying an Enterprise Vault/Consul Cluster for a customer, I discovered the following issue.

While setting up `vault_checksum_file_url` to the location where to find the Enterprise checksum for Vault Enterprise binary, which by default is set to `https://releases.hashicorp.com/vault/{{ vault_version }}/vault_{{ vault_version}}_SHA256SUMS` for OSS which is the file itself. I encountered a failure scenario because in `install_enterprise.yml` Ansible add at the end of this URL another `{{ vault_enterprise_shasums }}` value.

So I ended up having something like this, which is obviously failing :/

https://FQDN/.../vault/ent/1.1.0/vault-enterprise_1.1.0%2Bent_SHA256SUMS/vault-enterprise_1.1.0%2Bent_SHA256SUMS

So I presume we should better document what is expected in that variable, I mean not the location of the SHASUM but the first part of the URL, or update Ansible code to align with the way it works for Open Source. Just my 2 cents.

Hope that helps.

brianshumate:
Thanks for reporting this!

Sorry that you've had trouble with it, but I'll try to fix up the documentation to better describe what's going on wit that variable.

PR #130, opened 2019-11-23
WIP: Use the same process for enterprise and non-enterprise installs

jmariondev:
Right now enterprise installs use the entirely different install_enterprise.yml task file, which doesn't seem to work currently.

I just modified the normal install file with some conditionals to insert '-enterprise' where needed.

I've not yet tried a premium enterprise install, I think I need to switch the '+ent' to '+prem'?

Lastly, I'm not sure if Hashi is comfortable with the URL to the enterprise binaries being public, so I set the zip/checksum URLs to an empty string as a default. If this gets merged we should probably assert that the user has set a URL if they're using `vault_enterprise`.


jmariondev:
This would also fix #89.

brianshumate:
👋 This is a great idea, thanks!

It would definitely be awesome to also include `+prem` and `+pro` for those versions and I think leaving the binary zip/checksum URLs empty as you have would be fine so that users can specify the URLs they prefer and I am happy to add a note about this in the docs too!

PR #131, opened 2019-11-26
Add missing task to add vault group if enabled

karras:
This PR adds the missing task to create the `vault_group` if `vault_manage_group` is set to true. The variable was already present in the defaults but the corresponding task was missing.

The functionality is therefore aligned with the `brianshumate.consul` role. :)


brianshumate:
Cool, thank you!

PR #132, opened 2019-12-02
Fix default value in the README

ilpianista:
The default value for `validate_certs_during_api_reachable_check` has always been `true`.


brianshumate:
Thank you!

PR #138, opened 2019-12-19
fix: indentation of vault config file

rasta-rocket:
* new line between each block
* remove whites spaces before some block (eg awskms, ...)


bbaassssiiee (approved):
Thanks!

PR #139, opened 2019-12-27
Set tags vault_check to vault_status debug task 

NorthFuture:
If you skip vault health check Vault status will fail since check_result.status is not available


bbaassssiiee (approved):
Thanks!

PR #140, opened 2020-01-02
Add the possibility to enable vault commands to localhost

erbap:
After looking to the error "Error initializing: Put http://127.0.0.1:8200/v1/sys/init: dial tcp 127.0.0.1:8200: connect: connection refused" I created a listener for localhost that could be enabled by a variable.

The default behavior is like before, without a listener for localhost.


bbaassssiiee (changes requested):

bbaassssiiee (changes requested):
please use the variable "{{ vault_port }}"

bbaassssiiee:
Please rebase your branch to include the Github action checks.

bbayszczak:
Your TCP listener is probably listening on the instance IP address that's why you cannot reach it using `127.0.0.1`

With this use case, setting `VAULT_ADDR` to your instance IP should solve your issue without having to add a new Vault TCP listener

PR #142, opened 2020-01-07
Add support for RHEL 8

kwevers:
The package name is slightly different in RHEL 8.
With these minimal changes everything appears to work (tested in a non-clustered set-up)


brianshumate:
Thanks!

PR #119, opened 2019-09-25
🐛 Fix for #118 using command instead of shell module

bbaassssiiee:
ansible-vault/tasks/main.yml:19 -name: Check Vault installation

PR #146, opened 2020-02-29
Add raft backend initial support

securez:
I also commited a fix to a task that in my tests always report as chaged, this can be on another PR if you want.

### DIFF
--- a/templates/vault_backend_raft.j2
+++ b/templates/vault_backend_raft.j2
@@ -1,0 +1,4 @@
+storage "raft" {
+  path = "{{ vault_data_path }}"
+  node_id = "{{ vault_raft_node_id }}"
+}
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -65,6 +65,7 @@ vault_main_configuration_template: vault_main_configuration.hcl.j2
 vault_backend: consul
 vault_backend_consul: vault_backend_consul.j2
 vault_backend_file: vault_backend_file.j2
+vault_backend_raft: vault_backend_raft.j2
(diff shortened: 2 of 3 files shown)


bbaassssiiee:
Please omit the commit for `fix: Avoid changed task on version check`.

templates/vault_backend_raft.j2
```diff
@@ -0,0 +1,4 @@
+storage "raft" {
+  path = "{{ vault_data_path }}"
+  node_id = "{{ vault_raft_node_id }}"
+}
```
bbaassssiiee: Missing newline

jpiron:
For the version check I might have the same issue : https://github.com/ansible-community/ansible-vault/pull/119#issuecomment-602571715

templates/vault_backend_raft.j2
seb-braun: raft_node_id can't be the same in a multinode cluster.

securez:
Sorry for the delay, I'm a bit busy this days, I already removed the check fix from the pull request, I'll review #119 and create another one if required.

I alose corrected the missing space on template.

templates/vault_backend_raft.j2
securez: The spaces are fixed now. About **raft_node_id** this need to be diferent on every node/instace that can be specified via hosts_vars.

karras:
Is it possible to merge the Raft support soon? We'd love to deploy Raft clusters with this role as soon as possible. Thanks a lot for the great effort! :)

PR #149, opened 2020-03-15
Changed tls defaults according to documentation

### LINKED ISSUE #136: Default tls names differs from documentation
Hi,
great ansible role!
I figured out a small glitches in documentation / defaults/main.yml of roles vars.

- vault_tls_cert_file defaults to vault.crt
- vault_tls_key_file defaults to vault.key

in READMD.md these variables defaults are document as
- server.crt
- server.key

I personally would like the later as it would be consistent with your ansible-consul role.

regards
    Holger

zarath:
This would fix #136.

Attention. As this would change existing installations behaviour where people relay on settings from default. It propably shouldn't be committed in a minor version change.

Advantage of adjusting defaults is that then it is consistent to the consul role, which is often used in conjunction with this role, but in more widespread use.

Please tell me if you would prefer a MR which just modifies the README.md.


bbaassssiiee (approved):
Better

PR #150, opened 2020-03-15
Added user_groups variable

zarath:
Added the possibility to supply additional groups to the vault systems user.

Enables easier access to common used tls secrets without using the same group or user on all sharing services.


bbaassssiiee (approved):
Thanks!

PR #155, opened 2020-04-28
added role_name to match galaxy standards

ericsysmin:


bbaassssiiee (changes requested):
The role name is debatable now we have it in ansible-community. If we rename it however (to hashicorp-vault which is more appropriate INMHO) then we might break things. Having the name the same as the repo-name (ansible-vault) is more straightforward.

For now please change to:

```
role_name: ansible-vault
```

ericsysmin:
I'll just drop the PR then. Galaxy throws warnings when you have a `-` character in the role name, and even now it tries to convert it to lowercase

PR #163, opened 2020-06-05
Vault API reachable? Check DR Secondary

ghost:
In a DR setup, the role fails if ran against the nodes of the DR cluster. This change makes the task succeed even for DR secondary clusters.


bbaassssiiee:
Thanks!

PR #166, opened 2020-07-01
Recursively set TLS permissions

ilpianista:
When `vault_tls_config_path` already exists and `vault_tls_copy_keys` is set to `false`, only the certificates root folder (`vault_tls_config_path`) get the correct permission while the certificates inside aren't touched.

### DIFF
--- a/tasks/tls.yml
+++ b/tasks/tls.yml
@@ -4,12 +4,11 @@
 - name: Create TLS directory
   become: true
   file:
-    dest: "{{ item }}"
+    dest: "{{ vault_tls_config_path }}"
     state: directory
     owner: "{{ vault_user }}"
     group: "{{ vault_group }}"
-  with_items:
-    - "{{ vault_tls_config_path }}"
+    recurse: yes
 
 - name: Vault SSL Certificate and Key
   become: true
--- a/tasks/tls.yml
+++ b/tasks/tls.yml
@@ -8,7 +8,6 @@
     state: directory
     owner: "{{ vault_user }}"
     group: "{{ vault_group }}"
-    mode: "u=rwX,g=rX,o=rX"
     recurse: yes
 
 - name: Vault SSL Certificate and Key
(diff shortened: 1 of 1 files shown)


bbaassssiiee (changes requested):
Would setting the mode to 0700 complete this nice fix?

ilpianista:
> Would setting the mode to 0600 complete this nice fix?

Mmm, nope. `0600` isn't a valid mode for a folder.

ilpianista:
@bbaassssiiee Up

bbaassssiiee:
You're right! But let me rephrase please add a `mode:`, ansible-lint 4.3+ checks for that.

ilpianista:
Ok, it's a bit harder than that.

Since there's the `recurse` parameter, the `mode` that we specify are also applied to the files inside the folder, thus by making the files executable. As workaround I could use a symbolic mode (e.g. `u=rwX,g=rX,o=rX`), but that will overwrite the permissions which are already set on the files if `vault_tls_copy_keys` is set to `false`.

I'm not sure what's the best thing to do here.

bbaassssiiee:
Group and others don't need access to TLS files.

ilpianista:
> Group and others don't need access to TLS files.

Exactly. If I use `mode: "u=rwX,g=rX,o=rX"` combined with `recurse: yes`, then group and others get access to them.

ilpianista:
To recap:
- if we only set `recurse: yes`, then ansible-lint complains
- if we set `recurse: yes` && `mode: 0755`, then the TLS files get the executable bit
- if we set `recurse: yes` && `mode: u=rwX,g=rX,o=rX`, then the TLS files are readable by `group` and `others`.

May I simply add a `skip_ansible_lint` tag to this task?

bbaassssiiee:
Adding `skip_ansible_lint` is Ok if the `mode:` works.

bbaassssiiee:
Too outdated

ilpianista:
FTR: superseded by #306

PR #168, opened 2020-08-04
Allow local installation to be performed as a privileged user

mhitza:
Depending on the way the role is installed the local installation
steps will fail with a permission denied error.

Specifically I've encountered the issue in an Vagrant + Ansible Local
setup scenario, where the roles are installed using the pattern
described in the Vagrant docs:

https://archive.is/m8ZgO#install-galaxy-roles-in-a-path-owned-by-root

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,14 @@ Sorry, there is no planned support at the moment for Windows.
 
 The role defines variables in `defaults/main.yml`:
 
+### vault_privileged_install
+
+ - Set this to true if you see permission errors when the vault files are
+   downloaded and unpacked locally. This issue can show up if the role has
+   been downloaded by one user (like root), and the installation is done
+   with a different user.
+ - Default value: *false*
+
 ### `vault_version`
 
 - Version to install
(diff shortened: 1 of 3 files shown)


bbaassssiiee:
This role can be installed with [this Vagrantfile](https://github.com/dockpack/vault_dojo/blob/master/Vagrantfile)

mhitza:
> This role can be installed with [this Vagrantfile](https://github.com/dockpack/vault_dojo/blob/master/Vagrantfile)

You're using the ansible provider in that Vagrantfile, the issue is when using the ansible_local provider (roles are installed as root inside the VM)

bbaassssiiee (changes requested):
I guess this is OK for this use-case, but can you please introduce a variable for this behaviour?

mhitza:
@bbaassssiiee let me know how you'd like that variable to be called. I can't think of a good name.

I guess it would be a variable that defaults to `false` and it's interpolated/expanded in the `become` directives I have commented out in my previous PR.

bbaassssiiee:
`vault_privileged_install`

bbaassssiiee (approved):
Thanks!

PR #169, opened 2020-08-17
Add vault_systemd_service_name Variable

ghost:
This request adds a variable for the name of the systemd unit for vault.


bbaassssiiee (approved):
LFTM

PR #170, opened 2020-08-17
Change bashrc vault_addr

ghost:
This changes the .bashrc of the `vault_home` service user to point the `VAULT_ADDR` to `vault_address` and not `inventory_hostname`, because they need not be the same.


bbaassssiiee (approved):
Good point! Thanks!

PR #174, opened 2020-10-17
New option to install Vault via HashiCorp Linux Repo

zmingxie:
Introduce a new variable `vault_install_hashi_repo` which we can toggle to install Vault via HashiCorp Linux repository.

More info about this new repo can be found in this announcement: https://www.hashicorp.com/blog/announcing-the-hashicorp-linux-repository

### DIFF
--- a/tasks/install_hashi_repo.yml
+++ b/tasks/install_hashi_repo.yml
@@ -1,0 +1,32 @@
+---
+# File: tasks/install_hashi_repo.yml
+#       Install Vault via HashiCorp Linux repository
+
+- name: Add HashiCorp yum repo
+  command: yum-config-manager --add-repo=https://rpm.releases.hashicorp.com/RHEL/hashicorp.repo
+  args:
+    creates: /etc/yum.repos.d/hashicorp.repo
+  when: ansible_pkg_mgr == 'yum'
+
+- name: Install Vault via yum
+  yum:
+    name: "vault-{{ vault_version }}"
+    state: present
+  when: ansible_pkg_mgr == 'yum'
+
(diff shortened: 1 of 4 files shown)


tasks/install_hashi_repo.yml
```diff
@@ -0,0 +1,32 @@
+---
+# File: tasks/install_hashi_repo.yml
+#       Install Vault via HashiCorp Linux repository
+
+- name: Add HashiCorp yum repo
+  shell: yum-config-manager --add-repo=https://rpm.releases.hashicorp.com/RHEL/hashicorp.repo
```
bbaassssiiee: **ansible-lint**: 305: Use `shell:` only when shell functionality is required. Using `command:` is prefered.

bbaassssiiee (changes requested):
This could be a bit more consise using the generic `package:` module to install after configuring yum, dnf and apt with their specifics.

tasks/install_hashi_repo.yml
zmingxie: Fixed. Thanks for pointing it out!

bbaassssiiee (approved):

zmingxie:
> This could be a bit more consise using the generic `package:` module to install after configuring yum, dnf and apt with their specifics.

So I did try out using the `package:` module, but it doesn't seem to support installing a specific version of a package: https://docs.ansible.com/ansible/2.8/modules/package_module.html

That's why I'm using `yum` and `apt` separately here

bbaassssiiee:
Thanks for this contribution!

Issue #173, opened 2020-09-11
Downloading the bin everytime


rgarrigue:
Hi there

This one fails everytime, hence re-downloading the bin each time

```
TASK [brianshumate.vault : Check Vault installation] ******************************************************************************************************
fatal: [i-0ea3899fb8d4801ce]: FAILED! => {"changed": false, "cmd": "command -v vault", "msg": "[Errno 2] No such file or directory: b'command': b'command'", "rc": 2}
...ignoring
```

I'm running ansible 2.9.xx on Debian 10 if that helps. `command -v vault` works on my default bash shell, and sh too. I guess it's missing /usr/local/bin in its PATH.

zmingxie:
Did you override the `vault_bin_path`? It should be set to `/usr/local/bin` by default.

Also, which platform are you running into this issue?

rgarrigue:
Yes, the default path.

It's running on debian 10, on AWS EC2.

zmingxie:
You're right. I can reproduce this on Debian but not CentOS.

I'm not sure why this fails. Below is the full trace. Switching to `which` seem to fix this issue

```
TASK [ansible_vault : Check Vault installation] ***********************************************************
task path: /vagrant/roles/ansible_vault/tasks/main.yml:34
<localhost> ESTABLISH LOCAL CONNECTION FOR USER: vagrant
<localhost> EXEC /bin/sh -c 'echo ~vagrant && sleep 0'
<localhost> EXEC /bin/sh -c '( umask 77 && mkdir -p "` echo /home/vagrant/.ansible/tmp/ansible-tmp-1602956224.225932-261002391649915 `" && echo ansible-tmp-1602956224.225932-261002391649915="` echo /home/vagrant/.ansible/tmp/ansible-tmp-1602956224.225932-261002391649915 `" ) && sleep 0'
Using module file /usr/lib/python3/dist-packages/ansible/modules/commands/command.py
<localhost> PUT /home/vagrant/.ansible/tmp/ansible-local-7610_fcwo_gh/tmpe1nrve8h TO /home/vagrant/.ansible/tmp/ansible-tmp-1602956224.225932-261002391649915/AnsiballZ_command.py
[...]

PR #176, opened 2020-10-17
Update Vault installation check command

zmingxie:
This fixes the issue raised in https://github.com/ansible-community/ansible-vault/issues/173

'command' is not found in Debian favour linux which leads to new installation all the time.


bbaassssiiee:
which is not always installed. please test with `shell:` as suggested in the issue, and run ansible-lint too.

PR #179, opened 2020-10-21
feat(init) add vault init ability

eagleusb:
Add the ability to initialize Vault (cluster) when needed, it's a preview as I'll implement the output of unseal keys and the root token properly.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -63,6 +63,14 @@ vault_main_config: "{{ vault_config_path }}/vault_main.hcl"
 vault_main_configuration_template: vault_main_configuration.hcl.j2
 
 # ---------------------------------------------------------------------------
+# Initialization
+# ---------------------------------------------------------------------------
+
+vault_init: false
+vault_init_key_shares: 5
+vault_init_key_threshold: 3
+
+# ---------------------------------------------------------------------------
 # Storage backend
 # ---------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


bbaassssiiee (changes requested):
Secret-zero (operator init output) is not stored securely. Better implement this in a separate (project) role.

bbaassssiiee:
Maybe this inspires? https://github.com/dockpack/keybase_unseal

eagleusb:
Thanks for the inpu @bbaassssiiee . I'll add support for keybase and eventually GPG key for initialization.
To be honest I don't see how it will be more secured in another role, or copied from the table output manually.

The person deploying Vault with `ansible-vault` will (and want) initialize it, usually just copying the table output for unseal keys and root token. It would be a minimum to have that feature in `ansible-vault` IMO.

bbaassssiiee:
Not everything needs to go in one role. For comparison https://github.com/dockpack/vault_dojo uses this role and an couple of others to initialize vault create policies. Most of these thing depend on the organization using Vault, hence my objection to implement initialization in this role. This role is already quite complex.

Shamir secret sharing is one of the foundations of Vault. If you initialize and get unencrypted unseal keys, then that violates that principle.

eagleusb:
I know composition and it's a good engineering principle.

However the promise of `ansible-vault` is to deploy a Vault machine,cluster. Thus it has Raft, Consul... backends support and also auto-unsealing. 

There is a dependency error while you don't even support initialization, which is a necessary step for these. It's also an engineering principle violation, but I won't fight for it, as you said the role is quite complex and needs a refactorization.

Thanks for your time,
Regards

PR #180, opened 2020-11-16
Fix configuration setting for s3 backend paths.

lungj:
The `vault_backend_s3.j2` template has a setting for `vault_s3_force_path_style`. The actual setting's name is `s3_force_path_style` as per https://www.vaultproject.io/docs/configuration/storage/s3.


bbaassssiiee (approved):
LFTM

PR #185, opened 2020-12-10
become_user vault_user when modifying files in vault_home

### LINKED ISSUE #184: Exception due to permissions issue when modifying files in vault_home (ie. .bashrc)
Executing a playbook including this role as an `ansible` user. For the most part everything appears to be successful, but the play gets hung up on the `Insert http(s) export in dotfile` task since the `ansible` user doesn't have permission to modify the .bashrc file in the `vault` user's home directory:

```
TASK [ansible-community.ansible-vault : Insert http(s) export in dotfile] ******
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: OSError: [Errno 13] Permission denied: '/home/vault/.ansible_tmpIKn5bd.bashrc'
fatal: [vault-dev]: FAILED! => {"changed": false, "msg": "The destination directory (/home/vault) is not writable by the current user. Error was: [Errno 13] Permission denied: '/home/vault/.ansible_tmpIKn5bd.bashrc'"}
```

Would be nice if it was either possible to skip the `Insert http(s) export in dotfile` task, or maybe update it to run as the `vault` user to avoid permissions issues. Looking at the role there may be other tasks attempting to write to `vault_home` which would also be affected.

jrbeilke:
Fixes #184

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -244,6 +244,8 @@
   meta: flush_handlers
 
 - name: Insert http(s) export in dotfile
+  become: true
+  become_user: "{{ vault_user }}"
   lineinfile:
     path: "{{ vault_home }}/.bashrc"
     line: "export VAULT_ADDR='{{ vault_tls_disable | ternary('http', 'https') }}://{{ vault_address }}:{{ vault_port }}'"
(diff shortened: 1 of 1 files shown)


bbaassssiiee (changes requested):
Rather than become, add the owner and group properties to lineinfile actions. That is more portable.

bbaassssiiee (approved):

bbaassssiiee:
To fix #185 a tag could be introduced. 

> Would be nice if it was either possible to skip the Insert http(s) export in dotfile task,

PR #153, opened 2020-04-22
Service registration

Legogris:
This adds support for [service registration](https://www.vaultproject.io/docs/configuration/service-registration/consul) configuration.

This should resolve #152 .

PR #187, opened 2020-12-23
Fix consul/k8s service registration documentation and templates

DaazKu:
Both default value are missing the hcl extension.

https://github.com/ansible-community/ansible-vault/blob/42ae6c7e5a8990b08f72c6b9a300d548b3e11c47/defaults/main.yml#L165
https://github.com/ansible-community/ansible-vault/blob/42ae6c7e5a8990b08f72c6b9a300d548b3e11c47/defaults/main.yml#L185

The source of this mistake was an oversight on the PR #153 that introduced the service registration feature


bbaassssiiee (changes requested):
Thanks for spotting this. When I checked the files in templates I noticed that `templates/vault_service_registration_kubernetes.j2` is also missing that extension, and probably the `temlate:` action `src:` parameter needs to be adjusted as well to make this neat.

DaazKu:
> Thanks for spotting this. When I checked the files in templates I noticed that `templates/vault_service_registration_kubernetes.j2` is also missing that extension, and probably the `temlate:` action `src:` parameter needs to be adjusted as well to make this neat.

@bbaassssiiee There's no template/src because these files are included from the main template: https://github.com/ansible-community/ansible-vault/blob/42ae6c7e5a8990b08f72c6b9a300d548b3e11c47/templates/vault_main_configuration.hcl.j2#L53-L58

So the fix I just pushed should do it.

bbaassssiiee (approved):

bbaassssiiee (approved):

bbaassssiiee:
LFTM

PR #157, opened 2020-04-30
Adding Support of Raft Integrated Storage as Default Option

nehrman:

PR #193, opened 2021-01-13
Feature Toggle for Static/Dynamic Raft Node List

ghost:
This allows to choose between a statically, precompiled list of `retry_join` nodes to configure the Raft backend (`vault_backend: raft`) or a dynamically generated list that is generated using Ansible facts (`vault_backend: raft_dynamic`).

The reason for this (hopefully backward compatible) change is that having a dynamic option only (`vault_backend: raft_dynamic`) does not work for playbooks that run against a single node only, for instance, to perform an upgrade (batch size one).

The change https://github.com/ansible-community/ansible-vault/pull/157 requires Ansible to be aware of the facts of at least two other nodes in the group `vault_raft_group_name` (two, depending on your installation Ent/OSS).

In other words, the role fails at the following task, when run against just one node, having Raft backend enabled (excerpt):
```bash
$ ansible-playbook playbook.yml -i dev -e vault_raft_group_name=secondary -t ansible-vault -l vault3
...
TASK [ansible-vault : Vault main configuration] *******************************************************
fatal: [vault3]: FAILED! => {"changed": false, "msg":
"AnsibleUndefinedVariable: 'ansible.vars.hostvars.HostVarsVars object' has no attribute 'ansible_eth0'"}
```

Ansible tries to access the facts of the other nodes (e.g., `vault4` and `vault5`) and this information is not available because of `--limit vault3`.

The role fails, even when the primary address is added as a host variable in advance (`hostvars[server]['vault_api_addr']`).

This is certainly fine for initial installation, where we assume that the role is run against a proper group of at least 3 nodes.

However, we would like to re-use the role for upgrades. These upgrades should run in batches of size 1 ([`serial: 1`](https://docs.ansible.com/ansible/latest/user_guide/playbooks_strategies.html#setting-the-batch-size-with-serial)). This fails with the same error message as the example with `limit` above.

The `vault_raft_retry_join` variable used to specify the nodes in advance mitigates that problem. It is still explained in the Readme, but the code was removed. So let me propose to add this variable again?

Most likely, it will be necessary to integrate similar configuration snippets, for instance to configure [Raft backends and clusters in the cloud with auto join](https://learn.hashicorp.com/tutorials/vault/raft-storage-aws#cloud-auto-join)

Further changes in `vault_backend: raft_dynamic`:
* The role assumed that `vault_raft_group_name` is defined. Defaulting to `ansible_limit` might be a more sensible default (see example invocation above with `--limit`)? The assumption here is that we run the play for a whole host group of Raft nodes (`ansible_limit`, `-l`) for installation.
* Added check that `vault_raft_group_name` exists.
* Simplified if/else construct to include certificate files
* Removed `difference` in server to avoid template complexity

These suggestions intend to make the template file more readable, and as always, open to discussions.

Please let me know, which of the changes in the `vault_backend: raft_dynamic` template mentioned in the list above should be reverted to the current state in `master` again, if any.

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ The role defines variables in `defaults/main.yml`:
 
 ### `vault_backend`
 
-- Which storage backend should be selected, choices are: raft, consul, etcd, file, s3, and dynamodb
+- Which storage backend should be selected, choices are: raft, raft_dynamic, consul, etcd, file, s3, and dynamodb
 - Default value: raft
 
 
@@ -206,23 +206,6 @@ The role defines variables in `defaults/main.yml`:
 
 - CA certificate used for backend communication (if supported). This defaults to system bundle if not specified.
 - {{ vault_tls_ca_file }}
-
-### Raft Storage Backend
-
-#### `vault_raft_group_name`
-
-- Inventory group name of servers hosting the raft backend
-- Default value: vault_raft_servers
-
-#### `vault_raft_data_path`
-
-- Data path for Raft
-- Default value: vault_data_path
-
-#### `vault_raft_node_id`
-
-- Node_id for Raft
-- Default value: inventory_hostname_short
 
 ### Consul Storage Backend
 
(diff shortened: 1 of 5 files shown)


DaazKu:
You should probably look at https://docs.ansible.com/ansible/latest/user_guide/playbooks_delegation.html#delegating-facts for your use case.
IMO that would be better than complexifying the role.

ghost:
Thanks for the tip, this works perfectly.

PR #196, opened 2021-01-15
Fix systemd service and error redirection

### LINKED ISSUE #192: Vault cannot be stopped/restarted with systemd
Pre-conditions:
- os: Ubuntu 20.4
- logging turned on

I cannot properly stop or restart Vault with systemd because [it is executed in a subshell](https://github.com/ansible-community/ansible-vault/commit/a7b76466e1c2949c28f2db0d935d8a2e8a82d419#diff-2c0cdc895780005b16477aa338c276a01ba313763393e446b0d5609f92a1e017R33).
The command hangs util timeout is reached and the actual process is not stopped.

If I revert that back everything now works properly.

**Edit: Proper solution without using `StandardOutput/StandardError` #196**

----------------

In my case I was able to revert back and add the following under `[service]` instead.
```
StandardOutput=file:{{ vault_log_path }}/vault-output.log
StandardError=file:{{ vault_log_path }}/vault-error.log
```

The above is a feature of systemd v236 and [ubuntu 18.4 come with systemd v237](https://launchpad.net/ubuntu/bionic/+source/systemd) so that should be good on that front but Debian is a problem....
I don't know how to "fix" otherwise.

Here's the full template I ended up using:
```
### BEGIN INIT INFO
# Provides:          vault
# Required-Start:    $local_fs $remote_fs
# Required-Stop:     $local_fs $remote_fs
# Default-Start:     2 3 4 5
# Default-Stop:      0 1 6
# Short-Description: Vault agent
# Description:       Vault secret management tool
### END INIT INFO

[Unit]
Description="HashiCorp Vault - A tool for managing secrets"
Documentation=https://www.vaultproject.io/docs/
Requires=network-online.target
After=network-online.target
ConditionFileNotEmpty={{ vault_main_config }}
StartLimitInterval=60

[Service]
User={{ vault_user }}
Group={{ vault_group }}
ProtectSystem=full
ProtectHome=read-only
PrivateTmp=yes
PrivateDevices=yes
SecureBits=keep-caps
Capabilities=CAP_IPC_LOCK+ep
{% if systemd_version.stdout is version('230', '>=') %}
AmbientCapabilities=CAP_SYSLOG CAP_IPC_LOCK
{% endif %}
CapabilityBoundingSet=CAP_SYSLOG CAP_IPC_LOCK
NoNewPrivileges=yes
ExecStart={{ vault_bin_path }}/vault server -config={{ vault_main_config }} -log-level={{ vault_log_level | lower }}
ExecReload=/bin/kill --signal HUP $MAINPID
KillMode=process
KillSignal=SIGINT
Restart=on-failure
RestartSec=5
TimeoutStopSec=30
StartLimitIntervalSec=60
StartLimitBurst=3
LimitNOFILE=524288
LimitNPROC=524288
LimitMEMLOCK=infinity
StandardOutput=file:{{ vault_log_path }}/vault-ouput.log
StandardError=file:{{ vault_log_path }}/vault-error.log

[Install]
WantedBy=multi-user.target
```

P.S. logrotate would need to also be updated with the above since vault.err would also need to be rotated.

DaazKu:
Fixes https://github.com/ansible-community/ansible-vault/issues/192

Bash redirection are read from right to left. The error logs were never added to the file because the standard output was written before the error output was redirected to it.

Use `exec` to replace the parent process so that vault actually received the signal sent on stop/restart.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -154,7 +154,7 @@
 
 - name: "Set Exec output to log path when enabled log"
   set_fact:
-    vault_exec_output: "2>&1 >> {{ vault_log_path }}/vault.log"
+    vault_exec_output: ">> {{ vault_log_path }}/vault.log 2>&1"
   when: vault_enable_log
 
 - name: BSD init script
--- a/templates/vault_service_systemd.j2
+++ b/templates/vault_service_systemd.j2
@@ -39,7 +39,7 @@ Environment=HTTPS_PROXY={{ vault_https_proxy }}
 {% if vault_no_proxy -%}
 Environment=NO_PROXY={{ vault_no_proxy }}
 {% endif -%}
-ExecStart=/bin/sh -c '{{ vault_bin_path }}/vault server -config={{ vault_main_config }} -log-level={{ vault_log_level | lower }} {{ vault_exec_output }}'
+ExecStart=/bin/sh -c 'exec {{ vault_bin_path }}/vault server -config={{ vault_main_config }} -log-level={{ vault_log_level | lower }} {{ vault_exec_output }}'
 ExecReload=/bin/kill --signal HUP $MAINPID
 KillMode=process
 KillSignal=SIGINT


bbaassssiiee (approved):
By popular vote ;-)

PR #197, opened 2021-01-15
Added github workflow to auto-generate monthly tags

### LINKED ISSUE #195: Please resume making tagged releases
It appears that the last tagged release occurred when this project was taken over by the community, and the recommendation is now to use the master branch directly.

This is not ideal in a production environment, since some regressions (such as #192) are inevitable. If versioned releases were made regularly, even on an arbitrary schedule such as the 1st of every month, it would be much easier for users to revert to a previous version while bugs are fixed.

willhaines:
resolves #195

### DIFF
--- a/.github/workflows/monthly-snapshot.yml
+++ b/.github/workflows/monthly-snapshot.yml
@@ -1,0 +1,31 @@
+name: Release Monthly Snapshot
+
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


bbaassssiiee (approved):
Thanks! Let's see if this works.

PR #198, opened 2021-01-18
move StartLimitInterval, del StartLimitIntervalSec

patsevanton:
Unknown lvalue 'StartLimitInterval' in section 'Unit'
Unknown lvalue 'StartLimitIntervalSec' in section 'Service'

StartLimitInterval= and should be used [Service], not [Unit]
https://github.com/elastic/kibana/pull/47909/files
https://lists.freedesktop.org/archives/systemd-devel/2017-July/039255.html
_This switches our service file to use the older StartLimitInterval configuration. There have been a number of reports of startup errors due to older systemd versions not supporting the newer StartLimitIntervalSec config. The service files are backwards compatible, but not forwards compatible, so this moves back to the old format._

### DIFF
--- a/templates/vault_service_systemd.j2
+++ b/templates/vault_service_systemd.j2
@@ -14,7 +14,7 @@ Documentation=https://www.vaultproject.io/docs/
 Requires=network-online.target
 After=network-online.target
 ConditionFileNotEmpty={{ vault_main_config }}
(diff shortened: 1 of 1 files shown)


patsevanton:
Please review PR

bbaassssiiee (approved):
Thanks!

PR #201, opened 2021-01-19
Added support for useful options when running Vault behind a load balancer

willhaines:
Vault supports two methods for preserving the client IP when the Vault service is operating behind a load balancer or other type of reverse proxy. This change simply maps these configuration options through from Ansible variables to the vault configuration file.


bbaassssiiee (approved):
LGTM

PR #203, opened 2021-01-21
Separate seal config files

### LINKED ISSUE #202: Allow managing main configuration without providing secrets on every run
I will happily implement a PR, but starting with an issue to get input/feedback on this idea.

Currently, if you are using a cloud seal, such as AWS or Azure, you must provide the necessary secrets every time you run this role. This is not ideal if you want to use this role for performing upgrades and managing the rest of the Vault configuration.

You could store these secrets in an encrypted form with ansible-vault, however this has its own issues. What I'm proposing here is to separate out secret config values from non-confidential ones. This would allow these secrets to be provided only when specifically configuring them, otherwise they would simply be stored unchanged on the Vault hosts.

Vault supports reading multiple configuration files from the `vault.d` directory, so these could easily be separated without any complex merging logic.

willhaines:
This splits out the AWS KMS and Azure Key Vault configurations into separate .hcl files, so that `vault_main.hcl` may be managed without providing KMS credentials every time.

This also limits updating the Google KMS credentials file to only when a source file is specified, for the same reason.

Note that it was also necessary to change the Vault command to read the entire `vault.d` configuration directory, rather than just the `vault_main.hcl` file. In my opinion, this is the expected behavior for a *.d configuration directory.

Edit: Resolves #202

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -253,6 +253,7 @@ vault_tls_disable_client_certs: false
 
 # awskms seal
 vault_awskms: false
+vault_awskms_config: "{{ vault_config_path }}/vault_awskms.hcl"
 vault_awskms_backend: vault_seal_awskms.j2
 vault_awskms_region: "{{ lookup('env','AWS_DEFAULT_REGION') | default('us-east-1', false) }}"
 vault_awskms_access_key: "{{ lookup('env','AWS_ACCESS_KEY_ID') | default('', false) }}"
@@ -262,6 +263,7 @@ vault_awskms_endpoint: "{{ lookup('env','AWS_KMS_ENDPOINT') | default('', false)
 
 # azurekeyvault seal
 vault_azurekeyvault: false
+vault_azurekeyvault_config: "{{ vault_config_path }}/vault_azurekeyvault.hcl"
 vault_azurekeyvault_backend: vault_seal_azurekeyvault.j2
 
 # gcpkms seal
--- a/tasks/install_hashi_repo.yml
+++ b/tasks/install_hashi_repo.yml
@@ -33,6 +33,9 @@
 
 - name: Mask default Vault config from package
   copy:
+    owner: root
+    group: root
+    mode: "0644"
     dest: /etc/vault.d/vault.hcl
     content: |
       # Placeholder to mask default RPM/DPKG Vault config file.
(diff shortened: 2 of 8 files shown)


bbaassssiiee:
Rough idea: I think adding tags or extra-vars to certain actions can help, for instance to only run when wanted, of skipped in certain cases.

willhaines:
There were already variables to enable/disable the config sections that I moved into separate files, so I preserved that functionality. When we want to update our AWS KMS secrets, we run the following:

    export AWS_ACCESS_KEY_ID="<access key>"
    export AWS_SECRET_ACCESS_KEY="<secret access key>"
    export VAULT_AWSKMS_SEAL_KEY_ID="<key id>"
    
    ansible-playbook -K -e vault_awskms=true install.yml

To perform an upgrade or change other parts of the main config, we run the following (without the ENVs above set):
    
    ansible-playbook -K  install.yml

Where `install.yml` looks like this:

    ---
    hosts: vault_instances
    any_errors_fatal: true
    become: true
    become_user: root
    roles:
      - { role: brianshumate.vault, vault_backend: raft }


I'm certainly open to using tags as well, though. My current solution does not allow for deleting the `vault.d//vault_awskms.hcl` file.

willhaines:
If we added a tag, the interface could look something like this for AWS, and similar for Azure or Google.

To manage vault without touching KMS config:

    ansible-playbook -K  --skip-tags manage_kms install.yml

To update KMS secrets would be the same:

    export AWS_ACCESS_KEY_ID="<access key>"
    export AWS_SECRET_ACCESS_KEY="<secret access key>"
    export VAULT_AWSKMS_SEAL_KEY_ID="<key id>"

    ansible-playbook -K -e vault_awskms=true install.yml

To disable AWS KMS (deleting, or emptying `vault.d/vault_awskms.hcl`):

    ansible-playbook -K -e vault_awskms=false install.yml

I would prefer the opposite behavior  (have to explicitly enable a tag to manage KMS) to prevent accidentally wiping out the KMS config, but that would be a breaking change, and actually very easy for us to implement in our local `install.yml`.

bbaassssiiee:
Please rebase to include the molecule check in the PR.

bbaassssiiee (changes requested):
Please rebase to include the molecule check in the PR.

bbaassssiiee:
The checks fail.

bbaassssiiee (approved):

bbaassssiiee:
Thanks!

bbaassssiiee (commented):
Thanks for the heads up!

PR #207, opened 2021-01-28
Add vault_unauthenticated_metrics_access to listener

ghost:
Add support for [unauthenticated metrics access](https://www.vaultproject.io/docs/configuration/listener/tcp#configuring-unauthenticated-metrics-access)


bbaassssiiee (approved):
LFTM

PR #213, opened 2021-03-01
Remove init related info.

fbarbeira:
These blocks are not needed in a systemd service.


bbaassssiiee:
Thanks! LSB headers, nor chkconfig comments have meaning for systemd.

PR #216, opened 2021-03-18
Update VAULT_ADDR env var to be more TLS friendly

user404d:
If the vault_address value is 0.0.0.0 then the exported VAULT_ADDR URL inherits this ip addr. The IP addr being 0.0.0.0 is problematic if TLS certificates are used to communicate using mTLS when neither the server or client certificate include 0.0.0.0 as an IP SAN entry. A common pattern is to include 127.0.0.1 or localhost as SAN entries in TLS certificates so using the same swapping behavior as seen in the [Vault Reachable?](https://github.com/ansible-community/ansible-vault/blob/master/tasks/main.yml#L276) check seemed reasonable.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -247,10 +247,13 @@
 
 - name: Insert http(s) export in dotfile
   become: true
+  # Attempt to help with long lines > 160 issues
+  vars:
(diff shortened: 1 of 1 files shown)


bbaassssiiee (approved):
Looks fine, checks pass.

PR #218, opened 2021-03-27
Add 'vault_use_config_path' for out-of-band configurations

### LINKED ISSUE #219: Use 'vault_config_path' as config argument
Hello!

I ran into an issue yesterday where I was using cloud-init to add a configuration containing secrets at runtime; however, it was not getting picked up because the systemd vault config uses `-config={{vault_main_config}}` instead of `-config={{vault_config_path}}`.

I created a pull request #218 with I believe a backward-compatible way to do this but was wondering if this should be part of a larger depreciation effort as this seems to support a more general use case.

Thoughts?

kevinschoonover:
closes: #219

This PR introduces a couple of style changes that I'm happy to revert.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -60,6 +60,7 @@ vault_iface: "{{ lookup('env','VAULT_IFACE') | default(ansible_default_ipv4.inte
 vault_address: "{{ hostvars[inventory_hostname]['ansible_'+vault_iface]['ipv4']['address'] }}"
 vault_ui: "{{ lookup('env', 'VAULT_UI') | default(true, true) }}"
 vault_port: 8200
+vault_use_config_path: false
 vault_main_config: "{{ vault_config_path }}/vault_main.hcl"
 vault_main_configuration_template: vault_main_configuration.hcl.j2
 vault_listener_localhost_enable: false
--- a/README.md
+++ b/README.md
@@ -122,6 +122,11 @@ The role defines variables in `defaults/main.yml`:
 - Configuration file path
 - Default value: `/etc/vault.d`
 
+### `vault_use_config_path`
+
+- Use `"{{ vault_config_path }}"` to configure vault instead of `"{{ vault_main_config }}"`
+- default vaule: *false*
+
 ### `vault_plugin_path`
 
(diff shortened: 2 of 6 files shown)


defaults/main.yml
```diff
@@ -72,21 +73,21 @@ vault_no_proxy: ""
 # ---------------------------------------------------------------------------
 
 vault_tcp_listeners:
-  - vault_address: '{{ vault_address }}'
-    vault_port: '{{ vault_port }}'
-    vault_cluster_address: '{{ vault_cluster_address }}'
+  - vault_address: "{{ vault_address }}"
```
bbaassssiiee: I like these double quotes when there are "{{ moustache_variables }} or \n escapes",

defaults/main.yml
```diff
@@ -267,12 +268,12 @@ vault_azurekeyvault_backend: vault_seal_azurekeyvault.j2
 # gcpkms seal
 vault_gkms: false
 vault_backend_gkms: vault_seal_gcpkms.j2
-vault_gkms_project: ''
-vault_gkms_credentials_src_file: ''
-vault_gkms_credentials: '/home/vault/vault-kms.json'
-vault_gkms_region: 'global'
-vault_gkms_key_ring: 'vault'
-vault_gkms_crypto_key: 'vault_key'
+vault_gkms_project: ""
```
bbaassssiiee: use single quotes when strings are 'literal values' or values with potential [gotchas](https://docs.ansible.com/ansible/latest/reference_appendices/YAMLSyntax.html#gotchas)!

Empty strings should be single-quoted IMHO.

bbaassssiiee (changes requested):
Please create 2 PRs, one with the style-changes as suggested by my review. And after that gets merged create the one with the config path changes. Please create an issue for that change.

kevinschoonover:
@bbaassssiiee should be fixed now. Let me know if there are any other issues!

bbaassssiiee:
There are conflicting files now, are you able to resolve these?

bbaassssiiee (approved):

bbaassssiiee:
Thanks!

PR #221, opened 2021-04-28
Increase install speed

jtcarnes:
by using package's ability to use list (instead of calling package multiple times which with_items does) and not updating cache twice, we can speed up our code


bbaassssiiee:
Thanks!

bbaassssiiee:
Fix for #224

PR #222, opened 2021-04-28
Remove dependency on localhost for install

jtcarnes:
Many hosts have different proxies than their target systems. This causes issues if you try to use localhost to fetch everything for your targets and requires permissions on two systems (localhost and target) instead of just one.

By refactoring as such, you do not need to involve the local system at all, instead downloading vault directly to the target while still preserving checksums. This allows for host specific proxies and avoids issues of having things multiple places.


bbaassssiiee:
The variable `vault_install_remotely` is designed to solve that problem.

PR #225, opened 2021-04-29
Add Oracle Linux to Distribution Compatibility Check

ghost:
We were using this role to install Vault with [`vault_install_hashi_repo`](https://github.com/ansible-community/ansible-vault#vault_install_hashi_repo) on Oracle Linux 7. Can we add the "distro" (`ansible_distribution`) to return `true` for the compatibility check? Thanks


bbaassssiiee (approved):
The more, the merrier!

PR #227, opened 2021-05-20
rename task OS packages diff (Debian 9) as the task will also run on Debian 10

bbayszczak:


bbaassssiiee:
There was a merge issue, master should be OK.

PR #231, opened 2021-05-28
Add a pause after Vault starts

adrianmoisey:
I came across an issue where Vault would break it's state if it was
started, then immediatly restarted.

It seems as thought Vault didn't auto-unseal and join the raft cluster
before the restart happened.

To fix this, I added a pause to give Vault enough time to get going.


bbaassssiiee:
LFTM

PR #234, opened 2021-06-14
add `when` based on a var for handler `Restart vault`

bbayszczak:
Adding this condition will allow to choose if the playbook should be
able or not to restart the Vault service

Having this always on true on a production cluster could cause issues
because restarting all Vaults at the same time could end in a fully
sealed Vault cluster

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -4,3 +4,4 @@
 - name: Restart vault
   become: true
   service: name='{{ vault_systemd_service_name }}' state=restarted
+  when: vault_service_restart | bool
--- a/README.md
+++ b/README.md
@@ -40,6 +40,22 @@ specific software and versions:
 
 Sorry, there is no planned support at the moment for Windows.
 
+## Warning
+
+By default, this role may restart `vault` service when played (when there's a
+configuration change, OS Packages installed/updated)
+
+When there's no auto-unseal setup on your cluster, the restart may lead to a
+situation where all Vault instances will be sealed and your cluster will be
+down.
+
+To avoid this situation, the service restart by the playbook can be disabled
+by using the `vault_service_restart` role variable.
+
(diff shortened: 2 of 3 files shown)


bbaassssiiee:
Thanks for spotting this issue! Wouldn't `serial: 1` in the handler be better? Then they restart after eachother.

handlers/main.yml
```diff
@@ -4,3 +4,4 @@
 - name: Restart vault
   become: true
   service: name='{{ vault_systemd_service_name }}' state=restarted
+  when: vault_service_restart | bool
```
bbaassssiiee, line 7: Wouldn't serial: 1 in the handler be better? Then they restart after eachother.

bbaassssiiee (dismissed):

bbayszczak:
Adding `serial: 1` will still restart each Vault one after another almost instantaneously add the end of the playbook.
The  problem will remain the same, all Vaults will ends up sealed at the same time

I added this change to control a possible downtime

Example when changing a configuration where a restart is needed (with the change proposed):
1. Start the playbook which push the configuration change without restarting any instance
2. Then, restart and unseal all non master instances, this will not cause any downtime as the master is still receiving requests
3. Perform a `step-down` operation which will switch the leader to another instance (which is already restarted and already has the new configuration)
4. Restart and unseal the last instance which wasn't restarted as it was previously the master

Without this change, at the end of step 1, all Vault instances are restarted and sealed, even with the `serial: 1`

Aven adding `serial: 1` at the play level won't help either, you'll have to hurry to unseal the Vault instance that have been restarted before the next one is restarted :)

bbaassssiiee:
Maybe a note like this in the README?

bbayszczak:
> Maybe a note like this in the README?

Sure, what do you want to clarify ?

bbaassssiiee (approved):

Issue #211, opened 2021-02-12
AnsibleUndefinedVariable: 'ansible.vars.hostvars.HostVarsVars object' has no attribute ansible_eth0


fajarnugroho:
This error happened when using raft as storage backend.
```
TASK [ansible-community.ansible-vault : Vault main configuration] **********************
fatal: [server-1.company.internal]: FAILED! => {"changed": false, "msg": "AnsibleUndefinedVariable: 'ansible.vars.hostvars.HostVarsVars object' has no attribute 'ansible_eth0'"}
```

local machine details:
```
ansible-playbook 2.10.4
  config file = None
  configured module search path = ['/Users/nugrohof/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/Cellar/ansible/2.10.5/libexec/lib/python3.9/site-packages/ansible
  executable location = /usr/local/bin/ansible-playbook
  python version = 3.9.1 (default, Jan  8 2021, 17:17:43) [Clang 12.0.0 (clang-1200.0.32.28)]

Mac OS Big Sur 11.1
```

remote machine details:
```
Ubuntu 18.04.4 LTS
```

bbaassssiiee:
It's due to a concatenation, look at the ethernet interface name on your machine and define something like:
```
vault_iface: eth1
```

rkacg:
Hi,
i get the same error as fajarnugroho no matter if i omit or set the vault_iface parameter

**Error Message**
`TASK [ansible-vault : Vault main configuration] *****************************************************************************
fatal: [vault1.sec.austrocontrol.at]: FAILED! => {"changed": false, "msg": "AnsibleUndefinedVariable: 'ansible.vars.hostvars.HostVarsVars object' has no attribute 'ansible_ens160'"}`

**Server Interfaces:**
`[root@vault1 vault.d]# ip link
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
2: ens160: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP mode DEFAULT group default qlen 1000
    link/ether 00:50:56:ae:77:81 brd ff:ff:ff:ff:ff:ff`

**Playbok:**

hosts: vault
   gather_facts: True
[...]

PR #237, opened 2021-06-23
Revert "Replace server variable with inventory_hostname in raft

bbayszczak:
This change breaks the loop

It will also destroy any raft cluster with several nodes

For each Vault instance, there will be a list of X `leader_api_addr`
but always the same: the `inventory_hostname` of the current instance

This reverts commit b5e2f7b68af1025f451312d16a4c8d4986e5e4a9.

### DIFF
--- a/templates/vault_backend_raft.j2
+++ b/templates/vault_backend_raft.j2
@@ -19,14 +19,14 @@ storage "raft" {
   {% for server in groups[vault_raft_group_name] | difference([inventory_hostname]) %}
     {% if not (vault_tls_disable | bool) %}
   retry_join {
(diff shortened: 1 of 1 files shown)


bbayszczak:
@bbaassssiiee 
@adrianmoisey 

⚠️  Everyone using the role with this commit and a Raft cluster of multiple nodes will destroy his cluster

adrianmoisey:
Thanks, that does make sense. Sorry for the issue.

bbayszczak:
@adrianmoisey 
Be sure your host is a member of the `{{ vault_raft_group_name }}` group in your inventory
I hope it will solve your issue 🙂

adrianmoisey:
I've figured out the issue I was getting with https://github.com/ansible-community/ansible-vault/issues/211
I only got it when I limited to a single node in the group, which makes sense, as the other nodes won't have their data populated 🙈

bbaassssiiee:
Thanks for spotting this!

PR #238, opened 2021-06-23
add possibility to customize wait time after the first Vault start

bbayszczak:
Waiting 30 seconds for each playbook run shouldn't be a default
value as it's only impacting a few users.

It very time consuming especially for playbook runs with `serial`
and multiple nodes clusters (5 nodes: 2mn30s wait time)


bbaassssiiee:
Lftm

PR #239, opened 2021-06-23
add doc and fix vault_install_remotely

bbayszczak:
This option wasn't documented anymore

By default the playbook downloads the Vault binary on the controller host then push it to all targets

Setting this variable to `true` do the download directly from the targets


bbaassssiiee:
Thanks!

PR #240, opened 2021-06-24
add Raft cloud auto join settings

bbayszczak:
Add Raft cloud auto join settings

Hashicorp documentation can be found here
https://www.vaultproject.io/docs/configuration/storage/raft#retry_join-stanza


bbaassssiiee:
Thanks!

PR #243, opened 2021-07-15
Add Entropy Seal

ghost:
These changes adds the variable `vault_entropy_seal` (default: false). If set to True, it will include the [`entropy` seal stanza](https://www.vaultproject.io/docs/configuration/entropy-augmentation) in the main configuration file.


bbaassssiiee:
Thanks!

PR #244, opened 2021-07-17
fix: systemd unit option `ConditionFileExists`

ahjohannessen:
- `ConditionFileExists` option does not exist, it should be `ConditionPathExists`. See [unit docs](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#ConditionPathExists=).


ahjohannessen:
@bbaassssiiee I got this in systemd:

```
/etc/systemd/system/vault.service:6: Unknown key name 'ConditionFileExists' in section 'Unit', ignoring.
```

There is no `ConditionFileExists` option and changing to `ConditionPathExists` fixes it.

PR #246, opened 2021-07-23
Fix log rotate

ebdekock:
Log rotate doesnt currently work, once logs have rotated it still points
to the old file handler. This will ensure the logs function correctly once
it has rotated.


bbaassssiiee:
Thanks!

PR #247, opened 2021-08-04
Fix vault_seal_lib path in pkcs11 seal stanza

ghost:
The pipe to boolean will transform any string there to `False` when rendering the template.

This fix will render the original value of the `vault_seal_lib` path (i.e., the actual string).


bbaassssiiee (approved):
LFTM

PR #250, opened 2021-08-23
install: only install OS packages if non-empty

ahjohannessen:


ahjohannessen:
@bbaassssiiee WDYT? Flatcar and Fedora CoreOS do not have a package manager and this patch is to avoid running the package task when one of those OS are used.

bbaassssiiee (approved):
LFTM, Thanks!

PR #251, opened 2021-08-28
remote_install: only install OS packages if non-empty

ahjohannessen:
@bbaassssiiee WDYT? Same thing as with #250 (avoiding failure when using Flatcar or Fedora CoreOS)


ahjohannessen:
Thanks @bbaassssiiee  :+1: :)

Issue #252, opened 2021-10-01
Vault API reachable task fails when proxy env variables are defined


theodore86:
Hi,

At the end of the role the task **Vault API reachable?** could potentially fail when **http_proxy** or **https_proxy** environment variables have been set during the actual host configuration, is it feasible (?) to:

1. Set this task optional - using the ``when: {{ validate_vault_api | bool }}``
2. or for the ``uri`` module we could set the option ``use_proxy={{ use_proxy_during_api_reachable_check | bool }}``

I could implement it and perform an pull request.

```
# This should succeed regardless of seal state
- name: Vault API reachable?
  # Attempt to help with long lines > 160 issues
  vars:
    vault_addr_protocol: "{{ vault_tls_disable | ternary('http', 'https') }}"
  uri:
    validate_certs: "{{ validate_certs_during_api_reachable_check | bool }}"
    url: "{{ vault_addr_protocol }}://{{ vault_hostname | default(vault_addr, true) }}:{{ vault_port }}/v1/sys/health"
    method: GET
    # 200 if initialized, unsealed, and active
    # 429 if unsealed and standby
    # 472 if data recovery mode replication secondary and active
    # 473 if performance standby
    # 501 if not initialized
    # 503 if sealed
    # See: https://www.vaultproject.io/api/system/health.html
    status_code: 200, 429, 472, 473, 501, 503
    body_format: json
  register: check_result
  retries: 6
  until: check_result is succeeded
  delay: 10
  changed_when: false
  tags:
    - check_vault
```

Thanks,
Theo G.

bbaassssiiee:
A PR with option 2 would be nice...

PR #253, opened 2021-10-05
Optionally set proxy when Vault API reachable check

theodore86:
Add default boolean variable 'use_proxy_during_api_reachable_check' to
control whether proxy shall be used when vault api reachable task is
executed

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -328,6 +328,8 @@
   # Attempt to help with long lines > 160 issues
   vars:
     vault_addr_protocol: "{{ vault_tls_disable | ternary('http', 'https') }}"
+  environment:
+    no_proxy: "{{ vault_address }}"
   uri:
     validate_certs: "{{ validate_certs_during_api_reachable_check | bool }}"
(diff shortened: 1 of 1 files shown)


theodore86:
It seems that the **use_proxy** parameter requires to be accepted by ansible core: https://github.com/ansible/ansible/pull/74289

bbaassssiiee:
A workaround then?

environment:
  http_proxy: ''
  https_proxy: ''

or in uppercase..

theodore86:
Indeed, this is the only workaround at the moment to overcome any proxy configuration in conjuction with the ansible **uri** module. To "unset" the **http_proxy/https_proxy** environment variables at task level.

bbaassssiiee:
For the record, there is yet another workaround at the task level:

environment:
  no_proxy: "(( ip_of_host }}"

theodore86:
Indeed, better to set something like this:

```yaml
environment:
   no_proxy: "{{ vault_address }}"
```

bbaassssiiee:
Nice!

theodore86:
Hi,

Seems like i need an approval for the workflow to get started

bbaassssiiee (commented):
LFTM

bbaassssiiee:
Fixes #252

PR #261, opened 2021-11-23
Add a var to reload Vault instead of restarting it

akerouanton:
Some config changes can be taken into account by Vault with a reload. As restarting it might trigger an unseal process (eg. when it's sealed with Shamir keys), it might be quite disruptive. As such, this change adds a `vault_service_reload` var to tell the role to reload Vault instead. This is intended to be mostly set from the cli using: `$ ansible-playbook ... -e vault_service_reload=true ...` whenever needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,6 +218,11 @@ The role defines variables in `defaults/main.yml`:
 - Should the playbook restart Vault service when needed
 - Default value: true
 
+### `vault_service_reload`
+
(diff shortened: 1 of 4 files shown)


bbaassssiiee:
Could you please change the handler file to pure-yaml style?

akerouanton:
@bbaassssiiee Done :wink:

PR #263, opened 2021-11-23
Use vault_api_addr to set VAULT_ADDR in .bashrc

akerouanton:
Prior to this PR, `vault_api_addr` and `vault_cluster_addr` were defined using the IPv4 address of the default interface of the host. Moreover, a variable named `vault_hostname` was used to run the reachability check but was undocumented.

This change documents `vault_hostname` and leverages it to set the value of `vault_api_addr` and `vault_cluster_addr` when it's available. Otherwise, these vars are defined as before.

Moreover, `vault_api_addr` is now used to set the URL used by the reachability check and to set the value of `VAULT_ADDR` in .bashrc. Before that change, these URLs were define by concatenating the value of `vault_addr` and `vault_port` together. `vault_addr` was defined as being either `127.0.0.1` (when `vault_address` was `0.0.0.0`) or the value of `vault_address`. However, `vault_address` is used to define the IP address Vault should bind to. Although, by default, they were defined to be exactly the same as `vault_api_addr` default value, when overriding that param (for instance to put an hostname instead of the IP address), the value of `VAULT_ADDR` (and the URL used by the reachability check) could be wrong in some cases (eg. when using TLS certs with no `IP: 127.0.0.1` SAN). Instead of adding a new var to override the value of `VAULT_ADDR`, this change reuses the value of `vault_api_addr`, which is now defined through `vault_hostname` param (when provided).

### DIFF
--- a/README.md
+++ b/README.md
@@ -813,16 +813,24 @@ available starting at Vault version 1.4.
 - Address to bind to for cluster server-to-server requests
 - Default value: `"{{ hostvars[inventory_hostname]['ansible_'+vault_iface]['ipv4']['address'] }}:{{ (vault_port | int) + 1}}"`
 
+### `vault_hostname`
+
+- Hostname used to define `vault_cluster_addr` and `vault_api_addr` (when set).
+- Default value: none
+
 ### `vault_cluster_addr`
 
 - Address to advertise to other Vault servers in the cluster for request forwarding
-- Default value: `"{{ vault_protocol }}://{{ vault_cluster_address }}"`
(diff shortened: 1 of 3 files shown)


akerouanton:
`ansible-lint` breaks on `vault_api_addr` being too long but I don't know how to fix it. Writing it on multiple lines with `|-2` still adds a space between each parts. I'd need some help to fix this issue.

bbaassssiiee (changes requested):
please resolve conflicts

PR #264, opened 2021-11-26
Add vault_harden_file_perms to set chmod 0550 on config/plugins path

akerouanton:
The [Production Hardening](https://learn.hashicorp.com/tutorials/vault/production-hardening) have a bullet point "Allow minimal write privileges". It states: "its executable binary or any Vault configuration files".

Prior to this change, the config and plugins path had chmod 0750, so Vault could actually write config files and change plugins.

This commit adds a new parameter named vault_harden_file_perms (turned off by default). When enabled, it changes the chmod of config and plugins path to 0550 to effectively disallow Vault from writing into these dirs.

Side note: I hesitate between adding a new parameter and just setting chmod to 0550 and offering no way to configure that. The latter case is better because it's more secure but, although unlikely, I think it might break existing installation. I can update my PR if you think a new parameter is unnecessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,12 @@ The role defines variables in `defaults/main.yml`:
 - PID file location
 - Default value: `/var/run/vault`
 
+### `vault_harden_file_perms`
+
+- Whether this role should disallow Vault from writing into config and plugin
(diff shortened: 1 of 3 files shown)


bbaassssiiee:
Thanks!

PR #266, opened 2021-12-01
docs(readme): update available backend

Ameausoone:
add gcs and mysql


bbaassssiiee (changes requested):
, and more. For an up2date list refer to the documentation

https://www.vaultproject.io/docs/configuration/storage

PR #267, opened 2021-12-09
vault api health check via the actual vault api address

FalcoSuessgott:
After deploying vault a health check is running towards all vault nodes via their respective hostnames and ports:

```yaml
- name: Vault API reachable?
  # Attempt to help with long lines > 160 issues
  vars:
    vault_addr_protocol: "{{ vault_tls_disable | ternary('http', 'https') }}"
  environment:
    no_proxy: "{{ vault_address }}"
uri:
    validate_certs: "{{ validate_certs_during_api_reachable_check | bool }}"
    url: "{{ vault_addr_protocol }}://{{ vault_hostname | default(vault_addr, true) }}:{{ vault_port }}/v1/sys/health"
    method: GET
```

However, in enterprise environments it is common that the vault nodes are only accessible via a bastion host or jumphost and the vault API  is served through a frontfacing loadbalaner. Thus the task "Vault API reachable?" failed for me.

This PR changes the health check to `vault_api_addr` and additionally introduces a boolean flag `vault_disable_api_health_check` allowing the user to disable the final task in order to prevent the playbook from failing, if that is the case.

Let me know what you think :) 

Kind regards

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -351,7 +351,7 @@
     no_proxy: "{{ vault_address }}"
   uri:
     validate_certs: "{{ validate_certs_during_api_reachable_check | bool }}"
-    url: "{{ vault_addr_protocol }}://{{ vault_hostname | default(vault_addr, true) }}:{{ vault_port }}/v1/sys/health"
+    url: "{{ vault_api_addr }}/v1/sys/health"
     method: GET
     # 200 if initialized, unsealed, and active
     # 429 if unsealed and standby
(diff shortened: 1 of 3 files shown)


tasks/main.yml
```diff
@@ -351,7 +351,7 @@
     no_proxy: "{{ vault_address }}"
   uri:
     validate_certs: "{{ validate_certs_during_api_reachable_check | bool }}"
-    url: "{{ vault_addr_protocol }}://{{ vault_hostname | default(vault_addr, true) }}:{{ vault_port }}/v1/sys/health"
+    url: "{{ vault_api_addr }}/v1/sys/health"
```
bbaassssiiee, line 354: Why this change? Where is vault_api_addr defined?

FalcoSuessgott: right here sir:

```yaml
vault_api_addr: "{{ vault_protocol }}://{{ vault_redirect_address | default(hostvars[inventory_hostname]['ansible_'+vault_iface]['ipv4']['address']) }}:{{ vault_port }}"
```
https://github.com/ansible-community/ansible-vault/blob/master/defaults/main.yml#L125

bbaassssiiee (approved):
Thanks!

PR #268, opened 2021-12-27
Remove unused vault_group_name var

akerouanton:
Although this variable exists since the first ever commit, it's never
been used.


bbaassssiiee (approved):
Nice catch!

PR #269, opened 2022-01-20
Feature/allow static raft cluster members

markafarrell:
# Add feature to allow the manual configuration of Raft Cluster Members

## What?
Allows the static definition of members of the Raft cluster using the `vault_raft_cluster_members` variable.

e.g. 
  ```
  vault_raft_cluster_members:
    - peer: vault-host-1
      api_addr: https://vault-host-1:8200
    - peer: vault-host-2
      api_addr: https://vault-host-2:8200
    - peer: vault-host-3
      api_addr: https://vault-host-2:8200
  ```

Note: This variable defaults to the selecting hosts using the automatic method of selecting all hosts in the `vault_raft_group_name` group

Note: Also fixes setting `no_proxy` environment variable for API reachable check

## Why?
Currently Raft cluster membership is controlled using `vault_raft_group_name` which will dynamically select hosts that are members of the `vault_raft_group_name` group. This means, however, that the role must be run against all the hosts of a cluster at once.

The ability to run the role against single members of the cluster is important for staged rollout etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,6 +312,23 @@ vault_tcp_listeners:
 - Inventory group name of servers hosting the raft backend
 - Default value: vault_raft_servers
 
+### `vault_raft_cluster_members`
+
+- Members of the raft cluster
+- Default value: hosts in `vault_raft_group_name` group
+- Can be used to override the behaviour of dynamically selecting all hosts in `vault_raft_group_name`
+- Example:
+  ```
+  vault_raft_cluster_members:
+    - peer: vault-host-1
+      api_addr: https://vault-host-1:8200
+    - peer: vault-host-2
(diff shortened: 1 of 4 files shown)


FalcoSuessgott:
What about running ansible with the „—limit“ parameter in order to control which node should be configured?

markafarrell:
> What about running ansible with the „—limit“ parameter in order to control which node should be configured?

@FalcoSuessgott This issue with this is that the current approach requires host vars from the other members of the cluster which aren't accessible when you use `--limit`. When you use limit the playbooks can only access the hostvars of hosts included by the limit

lanefu (approved):

markafarrell:
@lanefu Is something missing here to get this reviewed?

PR #278, opened 2022-03-18
templates: backend raft: resolve to proper interface 

jimnydev:
This `for` loop is looping through all (remaining) hosts from group - but it's using `vault_iface` value of current host. I believe that's part of the issue with #211

There's also another change - instead of `['ansible' + vault_iface]` hostvars key - I've switched it to `ansible_facts` instead - as that one seems to be (always?) available. I'm not sure if presence of `'ansible_' + vault_iface` hostvars dict depends on ansible version, but either way - I've tried it on a couple of setups/installations and it seems to be passing everywhere.

Thanks.

### DIFF
--- a/templates/vault_backend_raft.j2
+++ b/templates/vault_backend_raft.j2
@@ -31,7 +31,7 @@ storage "raft" {
   {% for server in groups[vault_raft_group_name] | difference([inventory_hostname]) %}
     {% if not (vault_tls_disable | bool) %}
   retry_join {
-    leader_api_addr = "{{ hostvars[server]['vault_api_addr'] | default(vault_protocol + '://' + hostvars[server]['ansible_' + vault_iface]['ipv4']['address'] + ':' + (vault_port|string)) }}"
+    leader_api_addr = "{{ hostvars[server]['vault_api_addr'] | default(vault_protocol + '://' + hostvars[server]['ansible_facts'][hostvars[server]['vault_iface'] | default(vault_iface)]['ipv4']['address'] + ':' + (vault_port|string)) }}"
     {% if vault_raft_leader_tls_servername is defined %}
     leader_tls_servername = "{{ vault_raft_leader_tls_servername }}"
     {% endif %}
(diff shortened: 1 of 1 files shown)


kartuhn:
@jimnydev I still have the same issue even with the updated config file in your commit when running the test.

I added a VM under [vault_raft_servers] in test/inventory.ini file. Am I missing something? Thanks in advance for the help.

jimnydev:
Hosts should be listed within `[vault_raft_servers]` group, and if interface names differ across the hosts - you should specify `vault_iface` var for every host.

Which error are you seeing exactly? Thanks!

kartuhn:
@jimnydev Thanks. 

This is the tests/inventory.ini file. `raft` is just another ssh-able VM. 

```
localhost

[vault_raft_servers]
raft
```
The error that I'm getting, which I believe is related to #211 when running the main vault configuration task.

`{"changed": false, "msg": "AnsibleUndefinedVariable: 'dict object' has no attribute 'enp0s10'"}`

Could you please tell how to specify vault_iface for each hosts? Thanks in advance!

jimnydev:
In your example:

```
localhost

[vault_raft_servers]
raft vault_iface=interface_name_goes_here
```

kartuhn:
@jimnydev Thanks, I tried the following. 

```
localhost

[vault_raft_servers]
raft vault_iface=enp0s10
```
I can also confirm that interface on the localhost and raft in my example are the same using the `nmcli device status`

To add more context, this is how my tests/test.yaml looks like
```
---
- hosts: localhost
  connection: local
  remote_user: test
  become: true
  become_user: root
  roles:
    - ansible-vault
```
Could you please let me know if I'm missing something?

jimnydev:
Can you try this commit instead?

https://github.com/jimnydev/ansible-vault/commit/7368f4f4a3e815482c3598d9cce80eed2ee0f55e

bbaassssiiee:
Please resolve conflicts.

jimnydev:
It looks like this was reworked in the meantime - so this PR is now pointless.

Thanks!

PR #280, opened 2022-03-29
Add a param to customize apt distribution release

akerouanton:


bbaassssiiee (changes requested):
If you work on Debian/Ubuntu, then please add these to Molecule tests:

https://github.com/ansible-community/ansible-vault/blob/master/molecule/default/molecule.yml#L20

PR #281, opened 2022-03-29
Notify restart/reload handlers when TLS files change

akerouanton:


bbaassssiiee:
Unsealing the vault should be a standard operation.

PR #284, opened 2022-04-05
Allow overriding rpm repo url and support rhsm repo install

gardar:
- Allows overriding the repo url for rpm based distros too when previously only deb repos were supported (#259)
- Adds support for installing from RHSM repos (RedHat Satellite / Foreman / etc.)


bbaassssiiee (approved):
YOLO

PR #285, opened 2022-04-06
Simplify package checksum verification

gardar:
There's no need for get_url/shell/stat/etc. tasks when you can simply use a url lookup for fetching the package checksum.


bbaassssiiee:
Jinja2 Ninja! Thanks

PR #286, opened 2022-04-11
Feature request: Add HMAC key label to pkcs11 Seal stanza

ghost:
This feature request adds the `hmac_key_label` to the pkcs1 Seal stanza.

If not set to the empty string (`""`), the key label is used and set in the configuration file. This is the same behavior as for the `key_label`.

The HMAC key is required when using the default `0x1085` [mechanism](https://www.vaultproject.io/docs/configuration/seal/pkcs11#mechanism) (`CKM_AES_CBC_PAD`).

Done:
* The documentation was updated with the new key
* I changed the formatting using a "Jinja live parser", such that the pkcs11 seal stanza renders in a pretty format


ghost:
Could you give this a review @bbaassssiiee? Thanks

Issue #287, opened 2022-04-15
vault_disable_mlock should be available for all storage types


nununo:
Hello,

I'm trying to use this role to install Vault in a Proxmox LXC in a home lab context. Everything runs smoothly. Even the mlock capability check. But then the service fails to start with this error:

```
Apr 15 19:48:14 vault sh[2727]: Error initializing core: Failed to lock memory: cannot allocate memo>
Apr 15 19:48:14 vault sh[2727]: This usually means that the mlock syscall is not available.
Apr 15 19:48:14 vault sh[2727]: Vault uses mlock to prevent memory from being swapped to
Apr 15 19:48:14 vault sh[2727]: disk. This requires root privileges as well as a machine
Apr 15 19:48:14 vault sh[2727]: that supports mlock. Please enable mlock on your system or
Apr 15 19:48:14 vault sh[2727]: disable Vault from using it. To disable Vault from using it,
Apr 15 19:48:14 vault sh[2727]: set the `disable_mlock` configuration option in your configuration
Apr 15 19:48:14 vault sh[2727]: file.
```

This being a home lab I planned on disabling swap and mlock. But, while `vault_disable_mlock` is not mentioned in the documentation/defaults, is is being used in template `vault_backend_raft.j2`. Unfortunately, this is the only storage type accepting it.

Would it be possible to introduce `vault_disable_mlock` as a global configuration parameter available for all kinds of storages?

Thanks in advance.

nununo:
I just found a workaround which allows me to disable mlock using this role: using `vault_main_configuration_template` to replace the main template.

While I no longer need it, I still think it is a nice improvement so I'll leave the issue and the PR #288 open.

Regards.

PR #288, opened 2022-04-16
Make vault_disable_mlock a role parameter defaulted to false

nununo:
Proposal for implementing issue #287 which I posted earlier today.

### DIFF
--- a/templates/vault_main_configuration.hcl.j2
+++ b/templates/vault_main_configuration.hcl.j2
@@ -87,7 +87,7 @@ listener "tcp" {
   {% include vault_backend_raft with context -%}
 {% endif %}
 
-{% if vault_disable_mlock -%}
+{% if vault_disable_mlock is defined -%}
 disable_mlock = {{ vault_disable_mlock | bool | lower }}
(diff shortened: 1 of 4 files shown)


templates/vault_main_configuration.hcl.j2
```diff
@@ -87,6 +87,9 @@ listener "tcp" {
   {% include vault_backend_raft with context -%}
 {% endif %}
 
+{% if vault_disable_mlock -%}
```
bbaassssiiee: I'd expect `if vault_disable_mlock is defined`

nununo: Oh my! I believe you're right.

nununo: I just submitted a correction.

nununo:
The template `vault_main_configuration.hcl.j2` seems to have changed substantially since I submitted this PR. Hence the conflicts. I'm not sure the problem that this PR intends to solve is gone but, as it is, this PR is no longer relevant so I will close it.

PR #290, opened 2022-04-22
Add var vault_gkms_credentials_content

elcomtik:
To be able to provide Google KMS credentials by ansible var


elcomtik:
This is MVP, which I personally use. 

If the community like such idea, I would be glad to implement such a feature for other credential files managed by this role.

bbaassssiiee (commented):
LFTM

PR #292, opened 2022-05-04
fix(pkcs11): jinja whitespace

ghost:
@bbaassssiiee

This fixes a Jinja whitespace error in https://github.com/ansible-community/ansible-vault/pull/286.

The "live parser" I used was flawed and did not show the correct results when the role was applied.

This change should fix the whitespace errors.

Tested the following combination of values:
- With vault_softcard_enable=true
- With vault_softcard_enable=false
- With vault_seal_hmac_key_label=""
- With vault_seal_hmac_key_label="label"
- With vault_seal_key_mechanism=""
- With vault_seal_key_mechanism="0x1085"

Thank you for a review.

### DIFF
--- a/templates/vault_seal_pkcs11.j2
+++ b/templates/vault_seal_pkcs11.j2
@@ -1,17 +1,17 @@
 seal "pkcs11" {
   lib = "{{ vault_seal_lib }}"
-{%- if vault_softcard_enable %}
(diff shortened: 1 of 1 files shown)


ghost:
@bbaassssiiee may I request a short review on that request here? Thank you.

bbaassssiiee (commented):
LFTM

Issue #302, opened 2022-09-03
On RHEL8 or recent RedHat OSes libselinux-python should be replaced with python3-libselinux


dotdiego:
Can't install Hashicorp Vault using this role because of a missing package on most recent RedHat family OSes.

I'll create a pull request to resolve the issue using the solution in ansible-nomad role.

dotdiego:
It seems that Fedora36 doesn't automatically pulls the RedHat8.yml vars file

PR #303, opened 2022-09-03
fix: set vault_os_packages for RedHat

dotdiego:
#302

Is it possible to handle package installation conditionally like this ?
This method comes from ansible-community.ansible-nomad role.

I think that RedHat8.yml can be removed too with this solution.

### DIFF
--- a/vars/RedHat.yml
+++ b/vars/RedHat.yml
@@ -2,14 +2,6 @@
 # File: vars/RedHat.yml - Red Hat vars for Vault
 
 vault_os_packages:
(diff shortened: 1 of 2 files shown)


gardar:
Yes this can be done, but instead of having to define the name and version of every rhel based distro you can simply check if the system is running python 2 or 3.
Something like this:

```yaml
  - "{{ 'libselinux-python' if ansible_python_version is version('3', '<')  else 'python3-libselinux' }}"
```

And yes while you are at it, it would be good to delete the `vars/RedHat8.yml` file as it is no longer needed with this change.

dotdiego:
@gardar thanks for your input.
I've edited the file with your solution and removed RedHat8.yml

PR #305, opened 2022-09-30
Add support for Rocky distribution 

elcomtik:
This PR enables running role against Rocky distributions. I personally tested this role with Rocky8 and it works without issues.


bbaassssiiee (changes requested):
Please add Rocky to the molecule config

elcomtik:
I was wondering about the fact that the test exists only for Centos7. I'll rework the molecule code as it is in the ansible-consul repository and will add ones for Rocky 8. Maybe later I'll also add tests for other distros stated in meta.

bbaassssiiee:
Technical debt surfaces in the linter

elcomtik:
I'll fix that as well.

elcomtik:
Oh no, I see that it boils to https://github.com/ansible-community/ansible-vault/pull/155. Do you have a proposal on how to solve this? 

I see, that there is an ansible-lint skip_list for this, but it is ignored. I'll try to refactor it to use 'role-name' instead of deprecated '106'.

elcomtik:
That did not work. I'll set meta role_name to 'vault', to resolve this. 

Currently, the role is not published to the Ansible galaxy, so discussion about the role name may be put aside for later.

elcomtik:
It's failing at lint issues, which will be fixed by #308 , so I'll wait for it and rebase after it will be merged.

elcomtik:
I rebased it and removed my non-related work.

gardar (changes requested):
Looks good but minor nitpicking.
I think it would be preferable to have the molecule scenario named in a similar fashion to the others, `rocky-8` instead of `rocky8`... Or even call it `rockylinux-8` so that it's in line with the platform names.
Also don't forget to update the supported distros in the readme.

elcomtik:
> Looks good but minor nitpicking. I think it would be preferable to have the molecule scenario named in a similar fashion to the others, `rocky-8` instead of `rocky8`... Or even call it `rockylinux-8` so that it's in line with the platform names. Also don't forget to update the supported distros in the readme.

Of course, I will rename it & add info to README.

gardar (approved):

bbaassssiiee (approved):

bbaassssiiee:
Thanks!

PR #307, opened 2022-10-03
Stop concatenating spaces in templating logic.

deekayen:
The spaces before each template block are adding together to make the final product output file indenting excessive. Opting here for final readability instead of templating readability.

### DIFF
--- a/templates/vault_backend_raft.j2
+++ b/templates/vault_backend_raft.j2
@@ -1,51 +1,51 @@
 storage "raft" {
   path = "{{ vault_raft_data_path }}"
   node_id = "{{ vault_raft_node_id }}"
-  {% if vault_raft_performance_multiplier is defined and vault_raft_performance_multiplier %}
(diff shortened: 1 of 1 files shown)


gardar:
I think the indenting in the template might be there to make the template easier to read. 
It's a common practice to have the same indent level for the opening and closing tags of a jinja2 statement to be able to quickly realize where the statement begins/ends.
But I agree that it's definitely not something that should translate into the file that's generated from the template.

I wonder if the `lstrip_blocks` or `trim_blocks` parameters of the template module might fix the issue instead of removing the indentation from the template (and making it harder on the eyes).
It's also probably possible to fix the whitespace issue by using the minus sign ( `{%- %}` ) in some of the affected jinja2 statements, see: https://jinja.palletsprojects.com/en/3.0.x/templates/#whitespace-control

bbaassssiiee:
This branch has conflicts that must be resolved

lakrass:
I close this PR as it is very old.

PR #308, opened 2022-10-04
Molecule improvements: test supported platforms

gardar:
- Adds molecule scenarios for every platform (distro) that's currently supported by the role.
- Adds molecule tests for package (repo) install in addition to the zip install.
- Updated Github workflow which auto discovers all molecule scenarios so that when future scenarios/distros are added there's no need to touch the ci config.
- Few changes for how the molecule config is setup, to avoid repeats.
- Several fixes to pass ansible-lint and molecule tests.


elcomtik:
Nice work, I struggled with this in https://github.com/ansible-community/ansible-vault/pull/305 and it is already done here. I'll wait for the merge and rebase to continue with my work.

elcomtik:
@gardar my personal opinion is that using [action](https://github.com/robertdebock/molecule-action) from [robertdebock](https://github.com/robertdebock) is better because you only define matrix and one generic molecule config. You may see that in my PR https://github.com/ansible-community/ansible-vault/pull/305. 

Anyway thanks for fixing all of the lint issues :+1:

bbaassssiiee:
Thanks!

gardar:
> @gardar my personal opinion is that using [action](https://github.com/robertdebock/molecule-action) from [robertdebock](https://github.com/robertdebock) is better because you only define matrix and one generic molecule config. You may see that in my PR #305.
> 
> Anyway thanks for fixing all of the lint issues +1

Well actually you can define the matrix once and having a single molecule config in either one, I don't think it has anything to do with the molecule github action you use.
You can simply run `molecule test --all' which cycles through all of the scenarios. Or if you just have the default scenario then `molecule test` will suffice.

I personally think it's nicer to have multiple scenarios and then run each scenario in a separate job since then a single failure won't make your whole action fail, and you can simply re-run the failed job. It also means that you won't get all of the errors in a single log.

The scenarios only include configuration options that are relevant to that specific scenario, all of the default and shared options are located in the shared config under `.config/molecule/molecule.yml`. And you can actually overwrite any of those options by defining them in the scenario as you can see here:
https://github.com/ansible-community/ansible-vault/blob/f2be6f990719820b92bca3f0571cd702bbb35382/.config/molecule/config.yml#L20-L21
https://github.com/ansible-community/ansible-vault/blob/f2be6f990719820b92bca3f0571cd702bbb35382/molecule/archlinux/molecule.yml#L11-L13

elcomtik:
> Well actually you can define the matrix once and having a single molecule config in either one, I don't think it has anything to do with the molecule github action you use.

I  agree with you on this.

> I personally think it's nicer to have multiple scenarios and then run each scenario in a separate job since then a single failure won't make your whole action fail, and you can simply re-run the failed job. It also means that you won't get all of the errors in a single log.

That's not the case, you may look at how would it look in this action run https://github.com/robertdebock/ansible-role-dns/actions/runs/2986880975/jobs/4788931073

> The scenarios only include configuration options that are relevant to that specific scenario, all of the default and shared options are located in the shared config under .config/molecule/molecule.yml. And you can actually overwrite any of those options by defining them in the scenario as you can see here:

Currently, molecule scenarios are different only in 5 lines, specifically the names and images, which can be templated as you may see in https://github.com/robertdebock/ansible-role-dns/blob/master/molecule/default/molecule.yml

I have no intention to criticize or hate your work, just wanted to offer another solution to the same problem. I'm happy that tests are working now and we may further develop this wonderful role.

gardar:
> That's not the case, you may look at how would it look in this action run https://github.com/robertdebock/ansible-role-dns/actions/runs/2986880975/jobs/4788931073
 
> Currently, molecule scenarios are different only in 5 lines, specifically the names and images, which can be templated as you may see in https://github.com/robertdebock/ansible-role-dns/blob/master/molecule/default/molecule.yml

Ah clever! I understand it now, he's statically defining his job matrix and dynamically generating new scenarios whereas I statically defined the scenarios and then dynamically generate the job matrix.

I see pros and cons with both approaches, his approach can be more minimalistic, especially when the platforms and scenarios are all similar, but I can see it getting problematic when there are a lot of differences. 
Also you'd need a way to feed in the values from the matrix when running the molecule test outside of the ci.

> I have no intention to criticize or hate your work, just wanted to offer another solution to the same problem. I'm happy that tests are working now and we may further develop this wonderful role.

I didn't take it as a arguing or criticizing, it's good to have discussions and I'm always up to learning something new :smiley:

elcomtik:
Yeah, there are pros and cons. I didn't considered running molecule tests from local machine due to my minimalisic dev environment. 

I learned many new things also, so I'm happy with this discussion. We may let it go as current solution is good enough for this moment.

elcomtik:
@gardar look at comment for https://github.com/ansible-community/ansible-vault/pull/308/commits/1f200ad3c04178eaa9de0d0c5374e135c2ff64d2, I'm not sure if you get notify.

elcomtik (commented):
Please revert this because it breaks task. 

There is no templating, so if necessary add lint exclude.

tasks/main.yml
```diff
@@ -143,7 +143,7 @@
     - vault_gkms_credentials_src_file | length > 0
 
 - name: "Copy over GCP Credentials for Auto Unseal"
-  copy:
+  template:
```
elcomtik, line 146: Please revert this because it breaks task.

There is no templating, so if necessary add lint exclude.

gardar: Hmm, how does it break? Are you sure we can't fix it to function as a template? It would be better if we can fix it rather than adding a lint exclude.

elcomtik: Ok, I'll try to explain in detail.

At beginning the role was just copying 'vault_gkms_credentials_src_file' from control machine to managed host(s).

Later I implemented option to load this from variable in non breaking way. At first previous task check if there is file, if yes, then it will slurp its content into variable. 

So this variable can be set by inventory or slurped from ansible control machine.

In copy task we just take this content and copy it to managed host(s).

Ansible-lint is mistaken in detecting this as tempalte action. Moreover there is no elegant option to use variable in template module. Only option would by introduce template with just one variable inside it, what I think is serving to linter in way that is making code more  spaghetti then it should be.

elcomtik: I forgot to mention that template module doesn't have parameter content.

gardar: Ok got it, it's a bit ironic that the linter recommends using the template module and then doesn't complain that the content parameter is invalid for the template module.

As the template module always requires a file as a input I wonder if it would be good to use the lineinfile module to write the variable to a file? Since it can write the content of a variable to a file and it can create files if don't exist. We could then just use the regex in the lineinfile so that it will overwrite the whole file.

elcomtik: Using lineinfile could be the solution.

gardar: So I tried a few approaches, with using `lineinfile` `blockinfile` `replace` but was unable to get them to work correctly for this use case.
Only option left would be to create a template that just contains that single variable, but that's a silly solution imho.

gardar: Came across a recent discussion regarding this issue: https://github.com/ansible/ansible-lint/issues/2501

It looks like that in a future release of the linter this won't trigger a error, so until it gets released I think we should skip the lint on that task.

elcomtik: It looks that it gained a lot attention 😂

Thanks for investigation and fix.

PR #314, opened 2022-10-17
feat: add acme plugin

gardar:
Adds support for installing/enabling [vault-acme](https://github.com/remilapeyre/vault-acme/) plugin.

Depends on: #313 

Can be enabled with:

```yaml
---
vault_plugins_enable:
  - acme
```


gardar:
Managed to screw this branch up, rebased and opened a new PR #316

PR #318, opened 2022-11-23
fix: remove vault_datacenter

ghost:
This removes the unused variable `vault_datacenter`.

We don't understand the full history of this variable (why it was used initially), but it might originate from older installations where a Consul storage backend was required (the term `datacenter` makes more sense in the Consul realm).

The terminology in the Vault configuration is ["cluster name"](https://developer.hashicorp.com/vault/docs/configuration#cluster_name), which is already configurable with the `vault_cluster_name` variable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,11 +264,6 @@ The role defines variables in `defaults/main.yml`:
 - Cluster name label
 - Default value: dc1
 
-### `vault_datacenter`
-
-- Datacenter label
-- Default value:  dc1
-
 ### `vault_ui`
 
 - Enable vault web UI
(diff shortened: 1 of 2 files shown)


bbaassssiiee (changes requested):
Please remove debian-8 molecule scenario

bbaassssiiee:
I would expect checks to pass, did you rebase on master today?

ghost:
There appears to be an issue with the ACME plugin installation (`KeyError`). I had a quick look but was unable to fix it right away, needs more work

PR #321, opened 2022-12-08
fix(ci): failing legacy platforms and delegation issues.

gardar:
Fixes failing molecule ci jobs by: 
 - avoiding usage of omit in delegation since it seems to cause issues in some ansible versions.
 - downgrading the github runner from ubuntu-latest (which was recently bumped from 20.04 to 22.04) to ubuntu 20.04 since some of the legacy platforms supported by the role have systemd versions that are incompatible with the ubuntu 22.04 runner.

### DIFF
--- a/.github/workflows/molecule.yml
+++ b/.github/workflows/molecule.yml
@@ -13,7 +13,7 @@ env:
 
 jobs:
   discover-scenarios:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


FalcoSuessgott:
any thoughts on this @brianshumate ?

bbaassssiiee:
Brian is no longer involved in maintenance.

PR #325, opened 2023-03-02
Fix of template issue. User should be able to use TLS without his own CA

meklon-imhio:
Improved raft backend template: 
Added a condition to handle external non-controllable CAs like Let's Encrypt where we don't have access to vault_tls_client_ca_file, but still need TLS encryption for clients' connections in listeners. This should resolve any issues related to TLS encryption in such scenarios.

With this fix, TLS encryption can be activated for client connections in listeners even when using external CA certificates, ensuring secure communication.

### DIFF
--- a/templates/vault_backend_raft.j2
+++ b/templates/vault_backend_raft.j2
@@ -29,7 +29,7 @@ storage "raft" {
   {% endif %}
   {% if not vault_raft_cloud_auto_join_exclusive %}
   {% for raft_peer in vault_raft_cluster_members | rejectattr('peer', 'equalto', inventory_hostname) %}
-    {% if not (vault_tls_disable | bool) %}
+    {% if not (vault_tls_disable | bool) and vault_tls_client_ca_file != "" %}
   retry_join {
     leader_api_addr = "{{ raft_peer.api_addr }}"
     {% if vault_raft_leader_tls_servername is defined %}


meklon-imhio:
That's the desired configuration with Let's Encrypt certs. It's impossible with the original template, as the TLS activated adds the vault_tls_client_ca_file line to the storage "raft" block.
```
listener "tcp" {
  address = "555.222.333.444:8200"
  cluster_address = "555.222.333.444:8201"
  tls_cert_file = "/etc/vault.d/fullchain.pem"
  tls_key_file = "/etc/vault.d/privkey.pem"
  tls_min_version  = "tls12"
  tls_disable = "false"
  }

storage "raft" {
  path = "/opt/vault/data"
  node_id = "hasd-vault-de-1"
 retry_join {
    leader_api_addr =  "https://hasd-vault-nl-1.example.com:8200"
  }
 retry_join {
    leader_api_addr =  "https://hasd-vault-uk-1.example.com:8200"
  }
        }
```

meklon-imhio:
It looks like there are issues with almalinux test process:
`the role 'brianshumate.vault' was not found in...`

@brianshumate, could you please be so kind checking it? I think this error is not directly connected with my change.

FalcoSuessgott:
Yeah will try to look into that. You may want aswell it seems that linting also looks in example/ dir aswell which contains outdated files

FalcoSuessgott:
please rebase against master and trigger CI again

FalcoSuessgott:
I don’t understand why this change is necessary I will try to reproduce this and get back to you :)

meklon-imhio:
@FalcoSuessgott, thanks a lot :)

Please, let me clarify the idea a bit. The issue is connected with Letsencrypt + vault

1. You don't have your own CA and don't want to generate a self-signed one
2. You want to use encryption for your vault requests (sane idea even in controlled environment :) )
3. You've signed vault-node-1.example.com, vault-node-2.example.com for your purposes using certbot

The issue reveals when you're trying to turn on the encryption. When vault_tls_disable is false, then the template forces you to use this block:
```
  retry_join {
    leader_api_addr = "{{ raft_peer.api_addr }}"
    {% if vault_raft_leader_tls_servername is defined %}
    leader_tls_servername = "{{ vault_raft_leader_tls_servername }}"
    {% endif %}
    leader_ca_cert_file = "{{ vault_backend_tls_certs_path }}/{{ vault_backend_tls_ca_file }}"
    leader_client_cert_file = "{{ vault_backend_tls_certs_path }}/{{ vault_backend_tls_cert_file }}"
    leader_client_key_file = "{{ vault_backend_tls_private_path }}/{{ vault_backend_tls_key_file }}"
  } 
```
But you haven't got any vault_backend_tls_ca_file as it's on Let's Encrypt side.

With this fix you may use this block that is quite enough:
```
  retry_join {
    leader_api_addr =  "{{ raft_peer.api_addr }}"
  }
```
The TLS encryption would be turned on.

FalcoSuessgott:
Ok I understand the problem. One more question why did u condition it on vault_tls_client_ca_file?

meklon-imhio:
That was the easiest option I thought about.
With this fix we can have two options:

1. User wants just a general TLS encryption from client to vault node. He sets vault_tls_disable to false.
2. User wants total encryption and posesses a vault_tls_client_ca_file. He sets both of them.

Of course, we can add some kind of general trigger to set the desired encryption level. I just wanted to make as few changes as possible. Feel free changing it if you think that this solution doesn't suit well in the upstream, I'll adapt.

meklon-imhio:
Dear @FalcoSuessgott, could you be so kind having a look at this PR? It looks like all the tests passed successfully, so we can merge it in the base branch.

FalcoSuessgott:
LGTM, sorry for the delay

meklon-imhio:
Thank you very much for the attention. We'll migrate back to your upstream repository

PR #326, opened 2023-03-07
Retreive correct default iface dev name for raft members by default

karbyshevds:
Currently deployment is failing in case if vault nodes have different default network interface name.
E.x.:
- node1 have default network iface name ens160
- nodes node2 node3 have default network iface name wlp0s20f3

This PR makes default vault_raft_cluster_members work correctly when vault hosts have different device names.


FalcoSuessgott:
please rebase against master and trigger CI again

karbyshevds:
Rebased!

FalcoSuessgott:
Thank you!

PR #331, opened 2023-07-07
fix: shasum regex

ghost:
Extracting the shasum fails on my machine:
```
TASK [ansible-vault : Download Vault (local) → https://releases.hashicorp.com/vault/1.12.0+ent/vault_1.12.0+ent_linux_amd64.zip] ************************
fatal: [vlt-bootstrap-vlt-01 -> 127.0.0.1]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: No first item, sequence was empty.. No first item, sequence was empty.\n\nThe error appears to be in '***': line 32, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: \"Download Vault (local) → {{ vault_zip_url }}\"\n  ^ here\nWe could be wrong, but this one looks like it might be an issue with\nmissing quotes. Always quote template expression brackets when they\nstart a value. For instance:\n\n    with_items:\n      - {{ foo }}\n\nShould be written as:\n\n    with_items:\n      - \"{{ foo }}\"\n"}
```

Maybe it has to do with the fact that `vault_pkg` contains dots (`.`) which could mess up the regex? Not sure though.

In any case, testing for OS and arch should be sufficient, because the shasum file only contains hash sums for the same Vault version?

What do you think @gardar?

### DIFF
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -33,7 +33,7 @@
   get_url:
     url: "{{ vault_zip_url }}"
     dest: "{{ role_path }}/files/{{ vault_pkg }}"
-    checksum: "sha256:{{ (lookup('url', vault_checksum_file_url, wantlist=true) | select('match', '.*' + vault_pkg + '$') | first).split()[0] }}"
+    checksum: "sha256:{{ (lookup('url', vault_checksum_file_url, wantlist=true) | select('match', '.*' + vault_os + '_' + vault_architecture) | first).split()[0] }}"
     timeout: "42"
     mode: 0644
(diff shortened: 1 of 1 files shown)


ghost:
So seems like the original version works fine if I install from the OS repo with slightly changed variables:

```
vault_version: 'enterprise-1.11.2+ent'
vault_install_hashi_repo: true
```

The shasum check works fine when installing from the repo.

I will close this PR and go with that version (OS/repo package).

PR #332, opened 2023-07-07
feat: add vault_license_reporting var to opt out of automated license utilization reporting

ghost:
This adds a variable `vault_license_reporting` to allow enterprise customers to opt out of the [automated license utilization reporting](https://developer.hashicorp.com/vault/docs/enterprise/license/utilization-reporting).

The default for the variable is `false`, which means that no config changes are made and the license utilization is reported to HashiCorp automatically.

If the user sets the value to `true`, the license reporting is disabled in the main configuration file with the `reporting` stanza:
```
reporting {
    license {
        enabled = false
   }
}
```

Setting the value to `true` on Vault releases which do not yet support the reporting feature results in a warning in the Vault server log.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1269,6 +1269,12 @@ The role can configure HSM based instances. Make sure to reference the [HSM supp
 - Path to enterprise license on the Ansible controller (source file for upload). Upload skipped when empty or undefined. Only used if `vault_configure_enterprise_license: true`.
 - Default value: ""
 
+### `vault_license_reporting`
+
+- Opt out of automated License utilization reporting by setting this value to `false`. `vault_license_reporting=false` includes the necessary [`reporting`](https://developer.hashicorp.com/vault/docs/enterprise/license/utilization-reporting) stanza in the main configuration file.
+- The `reporting` stanza is only effective for Vault Enterprise >= `1.14.0`, Vault Enterprise >= `1.13.4`, Vault Enterprise >= `1.12.8`, Vault Enterprise >= `1.12.8` and  Vault Enterprise >= `1.11.12`. The `reporting` stanza is ignored (warning in the server log) in earlier Vault releases.
+- Default value: true
+
 ### `vault_hsm_app`
 
 - Set which cryptography app to use.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1269,6 +1269,12 @@ The role can configure HSM based instances. Make sure to reference the [HSM supp
 - Path to enterprise license on the Ansible controller (source file for upload). Upload skipped when empty or undefined. Only used if `vault_configure_enterprise_license: true`.
 - Default value: ""
 
+### `vault_license_reporting`
+
+- Opt out of automated License utilization reporting by setting this value to `false`. `vault_license_reporting=false` includes the necessary [`reporting`](https://developer.hashicorp.com/vault/docs/enterprise/license/utilization-reporting) stanza in the main configuration file.
+- The `reporting` stanza is only effective for Vault Enterprise >= `1.14.0`, Vault Enterprise >= `1.13.4`, Vault Enterprise >= `1.12.8`, Vault Enterprise >= `1.12.8` and  Vault Enterprise >= `1.11.12`. The `reporting` stanza is ignored (warning in the server log) in earlier Vault releases.
```
pree: Vault Enterprise `1.12.8` is double, also a doubled whitespace before Vault Enterprise >= `1.11.12`

templates/vault_main_configuration.hcl.j2
```diff
@@ -139,6 +139,16 @@ telemetry {
 license_path = "{{ vault_license_path }}"
 {% endif -%}
 
+{% if not (vault_license_reporting | bool) -%}
```
pree, line 142: As this is enterprise only, should we add `vault_enterprise` to this check, so it's not appended for OSS deployments?
I don't know if an OSS configuration fails if this is set.

pree (changes requested):

templates/vault_main_configuration.hcl.j2
ghost: Thanks for the review. Gave this scenario a test. We have the same behavior as with an Enterprise version which does not support reporting:
```
[WARN]  unknown or unsupported field reporting found in configuration at /etc/vault.d/vault_main.hcl:37:1
```

Apart from the warning Vault is fully functional.

README.md
ghost: Thanks for reviewing, fixed it

ghost:
I just noticed that we could probably achieve the same configuration by using the `vault_custom_configuration` value. However, adding the feature flag for this opt out is more explicit.

pree (approved):
LGTM :)

PR #336, opened 2024-01-31
restart according to rolling upgrade instructions added

smirta:
This will add a switch to restart a cluster according to instructions noted in [Rolling Upgrade Procedure for upgrading a single HA cluster](https://developer.hashicorp.com/vault/tutorials/standard-procedures/sop-upgrade#rolling-upgrade-procedure-for-upgrading-a-single-ha-cluster). 

Flow in ansible:
1. Identify if host is a leader
2. If not, restart it
3. Wait until it is available again
4. restart leader

I'm by no means an Ansible expert, I hope this makes sense and this PR is acceptable. As far as I can see it works as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,6 +279,11 @@ The role defines variables in `defaults/main.yml`:
 - Should the playbook restart Vault service when needed
 - Default value: true
 
+### `vault_rolling_service_restart`
+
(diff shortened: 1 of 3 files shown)


FalcoSuessgott:
thanks for the PR! Looks great, I will try to fix that CI very soon so you can continue

PR #342, opened 2024-06-24
feat: add Restart vault notification as in other install tasks

smirta:
Notify Restart vault handler as in the other install tasks.


smirta:
I accidently deleted the fork when cleaning up my profile.

lakrass:
In the meanwhile this change was already included in another PR.

PR #350, opened 2024-12-09
Add Debian 12 as molecule scenario and update role's meta information

sschmittsva:


sschmittsva:
Cannot get Debian 12 working. Will try again some other time.

PR #359, opened 2024-12-12
Changed when clause to use vault_tls_disable & vault_tls_gossip as bool instead of int comparison

lakrass:
- Cast variables to bool (in case someone uses env-vars and enters 0 or 1)
- Use bool query in when clause instead of integer comparison, so bool values can be assigned to the variables
- Changed default values of the variables to bool


lakrass:
This PR is ready for review!

FalcoSuessgott:
LGTM

lakrass:
Thank you! :)

PR #375, opened 2025-06-05
feat: Add option to bypass repository configuration

### LINKED ISSUE #374: Add option to bypass entire repository configuration
We need to host HashiCorp Vault in an airgapped environment on RHEL9 with artifactory as caching service for e.g. official RedHat- or Ubuntu-repos.
Right now the code will add the official public yum repos which are not reachable.
We handle our repos with a specific ansible role, so there is no need to add public repos.
I'm looking for a switch to bypass the repo-config by this role.

razorworks:
This fixes #374


lakrass:
Hey! The change itself makes sense! 
In the meantime the role has changed a bit. Please review and rebase your changes.

PR #376, opened 2025-06-12
 fix: solve crash if vault_port is set below 1024 by adding CAP_NET_ADMIN 

thomas-VIGINUM:
Add the possibility the possibility to set vault_port < 1024 without crashing by adding the `CAP_NET_BIND_SERVICE` capability when necessary.

### DIFF
--- a/role_variables.md
+++ b/role_variables.md
@@ -729,6 +729,8 @@ available starting at Vault version 1.4.
 ## `vault_port`
 
 - TCP port number to on which to listen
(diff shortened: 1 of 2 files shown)


lakrass (changes requested):
The change itself makes sense!

Please add proper documentation, so users are aware, that the additional capability is added when using a port < 1024.

Also please use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) and rebase your branch when finished.

thomas-VIGINUM:
hey ! is that ok for you ?

lakrass (approved):
This will work! :)
Thank you for your contribution!

PR #385, opened 2025-11-06
fix: allow user defined tls variables to be set instead of hardcoded default values

khawarizmus:
Currently using the role variables as such

```yml
vault_transit_tls_ca_cert: "vault-transit-ca-chain.crt"
vault_transit_tls_client_cert: "vault-autounseal.crt"
vault_transit_tls_client_key: "vault-autounseal.key"
```

Will totally ignore the values and will always yield the following config. the jinja template looks fine but the issue is that the variables are overridden by the defaults as you can see below

```hcl
seal "transit" {
  //... stuff here
  
  // TLS Configuration
  tls_ca_cert        = "/etc/vault.d/tls/supporting-ca-chain.crt"
  tls_client_cert    = "/etc/vault.d/tls/autounseal_client_cert.pem"
  tls_client_key     = "/etc/vault.d/tls/autounseal_client_key.pem"
}
```

This pr fixes this

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -330,8 +330,8 @@ vault_transit_key_name: 'autounseal'
 vault_transit_mount_path: "transit/"
 # vault_transit_namespace: ''
 vault_transit_tls_ca_cert_file: "{{ vault_tls_ca_file }}"
-vault_transit_tls_client_cert_file: "autounseal_client_cert.pem"
-vault_transit_tls_client_key_file: "autounseal_client_key.pem"
+vault_transit_tls_client_cert_file: "{{ vault_transit_tls_client_cert | default('autounseal_client_cert.pem', true) }}"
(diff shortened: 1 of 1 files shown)


lakrass (changes requested):
The changes itself lgtm! :)
Can you please reword your commit messages using [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) and rebase your branch? Then I'll approve the PR.

khawarizmus:
@lakrass I have done as requested. didn't realize that commit didn't follow convention.

lakrass (approved):

TobiX:
At first glance this didn't make any sense to me: Role defaults have lowest precedence, so this shouldn't be this complicated.

Upon further inspection this looks like a confusion between documented variable name (`vault_transit_tls_client_cert`, etc.) and implemented variable (`vault_transit_tls_client_cert_file`, etc.), but I guess it has been confusing since implemented (in 2022)...

I'll propose a fix to the documentation later...

Issue #398, opened 2025-11-28
new preinstall task fails


ahjohannessen:
After bumping version from v3.1.2 to v3.2.0 I get this:

```shell
TASK [ansible-vault : Update package cache] ***************************************************************************************************************************************************************************************************************************************************************************
task path: /Users/ahjohannessen/Development/Work/devops/ansible-libs/ansible-vault/tasks/preinstall.yml:21
redirecting (type: modules) ansible.builtin.atomic_container to community.general.atomic_container
[DEPRECATION WARNING]: community.general.atomic_container has been deprecated. Poject Atomic was sunset by the end of 2019. This feature will be removed from community.general in version 13.0.0. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
redirecting (type: modules) ansible.builtin.atomic_container to community.general.atomic_container
redirecting (type: modules) ansible.builtin.atomic_container to community.general.atomic_container
redirecting (type: modules) ansible.builtin.atomic_container to community.general.atomic_container
redirecting (type: modules) ansible.builtin.atomic_container to community.general.atomic_container
redirecting (type: modules) ansible.builtin.atomic_container to community.general.atomic_container
redirecting (type: modules) ansible.builtin.atomic_container to community.general.atomic_container
redirecting (type: modules) ansible.builtin.atomic_container to community.general.atomic_container
redirecting (type: modules) ansible.builtin.atomic_container to community.general.atomic_container
fatal: [192.168.81.101]: FAILED! => {"changed": false, "msg": "missing required arguments: backend, image, name"}
fatal: [192.168.81.102]: FAILED! => {"changed": false, "msg": "missing required arguments: backend, image, name"}
fatal: [192.168.81.100]: FAILED! => {"changed": false, "msg": "missing required arguments: backend, image, name"}
```

ahjohannessen:
ping @lakrass 

The task does not fail If the following in preinstall is removed:

```yaml
 - name: Update package cache
   ansible.builtin.package:
     update_cache: true
   tags: update_cache
```

The reason for this is that I use Fedora CoreOS in my staging env. ("msg": "atomic_container"). The same is going to happen for Flatcar Linux.

A fix is adding a condition:

```yaml

# in defaults:
  is_fedora_coreos: false
...
...
  when:
    - not is_fedora_coreos
    - not ansible_os_family in ["Flatcar", "Flatcar Container Linux by Kinvolk"]
```

or a more general thing:

```
--
[...]

PR #399, opened 2025-12-02
preinstall: account for fedora coreos

ahjohannessen:
@lakrass Not sure if you looked at the Fedora CoreOS issue I mentioned in #398 This fixes it, but perhaps a bit too "one-off".

### DIFF
--- a/tasks/preinstall.yml
+++ b/tasks/preinstall.yml
@@ -18,9 +18,32 @@
     system: true
   when: vault_manage_user | bool
 
+###
+
+- name: Check if booted with ostree
+  stat:
+    path: /run/ostree-booted
+  register: ostree
+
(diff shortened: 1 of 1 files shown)


lakrass:
Thank you for your PR! I think the #400 is a more elegant solution to the related issue.
Therefore I'll close this PR.

ahjohannessen:
Might work, Fedora CoreOS is Redhat variant, and packages in vars Redhad.ymt is not empty.

ahjohannessen:
@lakrass  Checked it. Does not work, as this is defined:

```
# File: vars/RedHat.yml - Red Hat vars for Vault
vault_os_packages:
  - "{{ 'libselinux-python' if ansible_python_version is version('3', '<') else 'python3-libselinux' }}"
  - git
  - unzip
```

PR #400, opened 2025-12-03
fix: only run cache update if we have packages to install

TobiX:
As an alternative to #399: Guard the package cache update with the same check we use to decide if we have packages to install. The later vault installation updates the cache anyways if a new repository is added to the package manager.

### DIFF
--- a/tasks/plugins/acme.yml
+++ b/tasks/plugins/acme.yml
@@ -10,15 +10,6 @@
     - name: Fetch acme vault plugin
       delegate_to: "{{ (vault_plugin_acme_install == 'local') | ternary('localhost', inventory_hostname) }}"
       block:
-        - name: Install dependencies
-          ansible.builtin.package:
-            name: "{{ vault_os_packages }}"
-            state: present
-          become: true
-          when:
-            - (vault_plugin_acme_install == 'remote')
-            - (vault_os_packages is defined) and (vault_os_packages | length > 0)
-
         - name: Create temporary directory for acme vault plugin
           ansible.builtin.file:
             path: "{{ (vault_plugin_acme_install == 'local') | ternary(vault_plugins_src_dir_local, vault_plugins_src_dir_remote) }}/acme"
--- a/tasks/preinstall.yml
+++ b/tasks/preinstall.yml
@@ -22,10 +22,11 @@
   ansible.builtin.package:
     update_cache: true
   tags: update_cache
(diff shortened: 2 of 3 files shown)


TobiX:
Test fail should be fixed by #401

ahjohannessen:
Fedora CoreOS is a Redhat ansible variant and default os packages is not empty. https://github.com/TobiX/ansible-vault/blob/1e2e639c87a26129077e03d833a80249cc4e955f/vars/RedHat.yml

ahjohannessen:
@TobiX This does unfortunately not fix the issue.

Ansible reports this for `Fedora CoreOS 42.20250929.3.0`:

`ansible_os_family: RedHat`
`ansible_distribution_major_version`: 42`

and 

```
- name: Include OS-specific variables
  ansible.builtin.include_vars: "{{ lookup('first_found', params) }}"
  vars:
    params:
      files:
        - "{{ ansible_os_family }}{{ ansible_distribution_major_version }}.yml"
        - "{{ ansible_os_family }}.yml"
      paths:
        - vars
```

loads:

```
# File: vars/RedHat.yml - Red Hat vars for Vault

vault_os_packages:
  - "{{ 'libselinux-python' if ansible_python_version is version('3', '<') else 'python3-libselinux' }}"
  - git
  - unzip
...
```

ending up with this error:

```
TASK [ansible-vault : Update package cache] ***************************************************************************************************************************************************************************************************************************************************************************
[DEPRECATION WARNING]: community.general.atomic_container has been deprecated. Poject Atomic was sunset by the end of 2019. This feature will be removed from community.general in version 13.0.0. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
fatal: [192.168.81.100]: FAILED! => {"changed": false, "msg": "missing required arguments: backend, image, name"}
```

ahjohannessen:
However, as I have a way overrule `vault_os_packages` I can work around it by setting this to empty.

lakrass:
I'll open a new issue to include this behavior by default. See #402

TobiX:
So installing packages would also not work since as an atomic OS the system doesn't have a package manager? If that is the case I find `ansible_os_family: RedHat` questionable at best... What's the value of `ansible_pkg_mgr`?

ahjohannessen:
The value of `ansible_pkg_mgr` is `atomic_container` - This is not true for Fedora CoreOS.

PR #404, opened 2026-01-02
Add dnf5 to list of package managers

jo5huajohn:
ansible_pkg_mgr in newer Fedora distributions report dnf5 instead of dnf. In this role, this prevents the `Add Vault/Hashicorp rpm repo` task from installing the Hashicorp repo, which causes dnf to attempt to install openbao with the version 1.21.1 which does not exist.


lakrass:
Looks good, thank you!

PR #409, opened 2026-02-18
feat: provide ability to configure missing mysql storage options

haegar:
Seeing that the mysql storage should be able to provide HA support for vault, and us already having a fully redundant MariaDB Galera/Maxscale cluster setup available, I wanted to try using it with vault - but found that the "ha_enabled" option was missing from the ansible role.
Thus adding this and the rest of the documented but missing settings with this PR.


lakrass (dismissed):

lakrass (approved):
Somehow the checks are not running but the code looks fine

PR #412, opened 2026-03-20
Leader CA can be empty to fall back to system certs

danishk-bp:


templates/vault_backend_raft.j2
```diff
@@ -35,7 +35,9 @@ storage "raft" {
     {% if vault_raft_leader_tls_servername is defined %}
     leader_tls_servername = "{{ vault_raft_leader_tls_servername }}"
     {% endif %}
+    {% if vault_backend_tls_ca_file | default('') != '' %}
```
lakrass, line 38: Please stick to the notation of the other if clauses: "if ... is defined"

lakrass (changes requested):

PR #419, opened 2026-04-21
BREAKING CHANGE: Enterprise + Hardening Updates

drewmullen:
Summary:

Proposing some changes per [Vault's Production Hardening Guide](https://developer.hashicorp.com/vault/docs/concepts/production-hardening)

1. Optionally disable shell history
2. Optionally disable swap
3. Optionally add version lock to users who install via `dnf`. This is useful when someone wants to manage vault upgrades outside of this role.
4. Optionally include `reporting {}` block in main config

All default behaviors are same as today (aka nothing is disabled or included without setting vars)

### Reporting block details

Vault enterprise, by default, reports license usage metrics to HashiCorp. This can be [optionally disabled](https://developer.hashicorp.com/vault/docs/configuration/reporting) via configuration. This PR allows users to disable or configure aspects of the reporting block.

To keep the template clean, I added the inclusion logic above via `set`. This is slightly different from convention but makes reading the stanza much cleaner. The `timestamp` value is the only exception, with its own nested `if` statement because that value has no true default and should be omitted unless specified.

To test without a full sweet run this playbook:

```yml
- hosts: localhost
  gather_facts: false
  vars:
    vault_cluster_name: "dc1"
    vault_max_lease_ttl: "768h"
    vault_default_lease_ttl: "768h"
    vault_cluster_disable: "false"
    vault_cluster_addr: "https://127.0.0.1:8201"
    vault_api_addr: "https://127.0.0.1:8200"
    vault_plugin_path: "/usr/local/lib/vault/plugins"
    vault_tcp_listeners: []
    vault_listener_localhost_enable: false

    vault_backend: file
    vault_backend_file: vault_backend_file.j2
    vault_data_path: /var/vault

    vault_service_registration_consul_enable: false
    vault_service_registration_kubernetes_enable: false
    vault_ui: true
    vault_entropy_seal: false
    vault_enterprise_hsm: false
    vault_gkms: false
    vault_ocikms: false
    vault_telemetry_enabled: false
    vault_configure_enterprise_license: false

    # scenario under test
    vault_enterprise: true
    vault_reporting_snapshot_retention_time: "9600h"
    vault_reporting_disable_product_usage_reporting: true
    vault_reporting_license_enabled: true
    vault_reporting_license_billing_start_timestamp: ""
    vault_reporting_license_development_cluster: false

  tasks:
    - ansible.builtin.template:
        src: templates/vault_main_configuration.hcl.j2
        dest: /tmp/vault_main.hcl
```

run playbook:

```shell
ansible-playbook -i localhost, -c local render-config.yml
```

check output - will show stanza if included:
```shell
grep -nE '^(reporting|  snapshot_retention_time|  disable_product_usage_reporting|  license|    enabled|    billing_start_timestamp|    development_cluster)' /tmp/vault_main.hcl || true
```

^ you can change `vault_enterprise = false` to verify its not included


defaults/main.yml
```diff
@@ -418,6 +421,14 @@ vault_license_file: ""
 # Upload skipped when empty or undefined, if `vault_license_file` is also empty or undefined
 # Only used if `vault_configure_enterprise_license: true`
 vault_license_content: ""
+# reporting stanza settings (enterprise only)
+# Added to config only if `vault_enterprise: true` and one or more values are non-default
+# https://developer.hashicorp.com/vault/docs/configuration/reporting
+vault_reporting_snapshot_retention_time: "9600h"
+vault_reporting_disable_product_usage_reporting: false
```
lakrass, line 428: Can you please turn off (vault_reporting_disable_product_usage_reporting = true) by default?
Sending data to a third party should be an optional feature.

lakrass (changes requested):

defaults/main.yml
lakrass: Everything else looks good :)

drewmullen: It’s currently set to how the tool defaults. Better to leave as is so any current users do not unexpectedly stop reporting usage which could have license issues. 

The PR was written in such a way so the block does not even appear unless a default value is changed. 

Said another way - anyone who deploys w enterprise = true today is already reporting usage, regardless of this change. It was previously impossible to disable reporting- which this PR now allows

lakrass: Ok got it! Didn't knew that Vault Enterprise already has been sending usage. 
In that case the PR is fine this way!

lakrass (approved):

drewmullen:
For the record - I dont think anything is breaking from this change. Everything added is default disabled or the same as before! <3 thank you for your attention and merging

PR #423, opened 2026-05-22
feat!: support Vault-based logging and remove old logrotate

hoangphuocbk:
Since HashiCorp Vault supports native logging, use Vault's built-in configuration to manage logging instead of logrotate.
https://developer.hashicorp.com/vault/docs/commands/server#_log_format


lakrass (commented):
LGTM, but due to removal of the `vault_exec_output` this will be a breaking change

PR #427, opened 2026-07-01
feat: allow to configure `unauthenticated_metrics_access` for each tcp listener

alesharik:
Hi,
this small PR adds ability to configure `unauthenticated_metrics_access` for each TCP listener separately. This is useful when someone wants to expose vault directly on some TCP targets, but keep `unauthenticated_metrics_access` enabled only for local metrics collector (which should use different TCP listener, which listens only on localhost/internal network)


lakrass (approved):
LGTM

lakrass:
Hey @alesharik, thank you for contributing! :)
Please fix the listing errors to make this PR mergeable.

alesharik:
Hi @lakrass ,
reworded problematic commit - now lint should pass

Issue #431, opened 2026-07-28
support Vault Agent Process Supervisor for Vault Server lifecycle


sgabenov:
### Summary

Add an opt-in mode that lets this role install and configure Vault Server without creating or managing its init/systemd service.

This enables Vault Agent Process Supervisor to own the Vault Server lifecycle, including starting Vault only after required secrets are available and restarting it when those secrets change.

### Why this matters

Vault Agent Process Supervisor is an official Vault capability for launching a child process with secrets injected as environment variables:

https://developer.hashicorp.com/vault/docs/agent-and-proxy/agent/process-supervisor

A high-value use case is transit auto-unseal.

Vault Server needs credentials to access the external Vault transit seal. A deployment should be able to use Vault Agent auto-auth to acquire and renew these credentials dynamically, rather than persist a long-lived transit token in:

- Ansible inventory or CI variables;
- rendered Vault Server configuration;
- systemd unit files;
- manually managed environment files.

In this model, Vault Agent waits until credentials are rendered, starts `vault server` as its child process, and can restart the child process when injected credentials are renewed or rotated.

This improves credential lifecycle management and reduces the amount of sensitive bootstrap material handled directly by deployment automation.

### Current limitation

The role currently always creates and starts the Vault init/systemd service.

That conflicts with process supervisor mode because systemd and Vault Agent would both attempt to manage the same Vault Server process. It may also start Vault before Vault Agent has authenticated and rendered the credentials required by the server.

### Proposed change

Introduce a new variable:
```yaml
vault_agent_supervisor_mode: true
```

Default: false.
When set to true, the role continues to:
- install Vault and required packages;
- create users, directories, TLS and storage configuration;
- render Vault Server and seal configuration;
- apply existing hardening and installation logic.

But it does not:
- create BSD, SysV, Debian init, or systemd service units;
- reload systemd;
- enable or start Vault;
- restart or reload Vault through role handlers.

This keeps the current behaviour unchanged by default and makes the role compatible with externally managed Vault Server processes.
Example:

```
- hosts: vault_servers
  roles:
    - role: community.vault
      vars:
        vault_agent_supervisor_mode: true
        vault_transit: true

        # Use this when Vault Agent Supervisor starts Vault after the role
        # has finished rendering its configuration.
        vault_disable_api_health_check: true
```

Vault Agent configuration, auto-auth method, policies, and token delivery remain outside this role’s scope.
Implementation

[...]

PR #432, opened 2026-07-28
feat: support Vault Agent Process Supervisor mode

sgabenov:
## Vault Agent Process Supervisor mode

Adds support for running Vault Server under Vault Agent Process Supervisor through the new `vault_agent_supervisor_mode` variable (default: `false`).

When enabled, the role still installs Vault and renders its configuration, but does not create or manage a Vault init/systemd service: it does not enable, start, reload, or restart Vault.

This lets Vault Agent Process Supervisor own the Vault Server lifecycle, particularly for transit auto-unseal without exposing a long-lived token through Ansible variables or a systemd unit.

Adds a Molecule scenario that verifies no role-managed systemd unit is created in this mode.

Related upstream issue: https://github.com/ansible-community/ansible-vault/issues/431

## Migration note

When enabling supervisor mode on a host previously managed by this role, stop and disable the old service before the handoff. The role intentionally leaves an existing service untouched so it cannot stop Vault before the supervisor is ready.

The Molecule scenario verifies that the package-provided `vault.service` remains disabled and stopped and that no role-managed unit is created.


sgabenov:
@lakrass is it possible to review a PR?

sgabenov:
Update after review: synchronized the branch with current master, documented the service handoff required when migrating an already-managed host, and strengthened the Molecule checks so the package-provided vault.service must remain disabled and stopped in supervisor mode. The role syntax check and both rendered Goss variants pass locally. @lakrass, could you please approve CI and review the update?

lakrass (approved):

lakrass:
Code looks good as well here!
Thank you for your contributions!

lakrass:
Same for the Debian 11 test here: the os is not supported anymore, so failing test is fine for now

